PR #10, opened 2025-07-18
added mcp deploy functionality for ts/py single-file apps

masnwilliams:


src/app/[transport]/route.ts
```diff
@@ -692,6 +703,266 @@ const handler = createMcpHandler((server) => {
       }
     },
   );
+
+  // Deploy App Tool
```
Comment (rgarcia) [inline src/app/[transport]/route.ts:707]: couple of thoughts reading this:

- should we let this tool accept an array of files in case it make sense to factor things out that way 
- the reason I ask is I wonder if you really started iterating on an app with this if you'd want to expose a "edit_app" tool that just feeds diffs or new files

We can see how it goes for now but want to make sure the experience of iterating on an app is somewhat quick / smooth

rgarcia (dismissed):
so cool!

src/app/[transport]/route.ts
Comment (masnwilliams) [reply src/app/[transport]/route.ts:707]: Added multi-file support. includes nested/flat directories

PR #104, opened 2026-05-24
add managed auth tools (manage_auth_connections, manage_credentials, manage_credential_providers)

masnwilliams:
## Summary

Closes the largest agent-facing capability gap in the MCP server: setting up an authenticated browser session for a third-party site. Agents can now drive Kernel's managed auth flow end-to-end without a human in the loop beyond the one-time hosted login or explicit credential handoff.

## What's added

- **`manage_auth_connections`** (full surface)
  - `create` — start managing auth for a profile + domain (optionally referencing a pre-stored credential by name, or an external provider like 1Password)
  - `list` / `get` / `delete`
  - `login` — kicks off a hosted login flow. Returns `hosted_url` (share with the user to sign in) and `live_view_url` (agent can watch). Triggers automatic re-auth if credentials are saved.
  - `submit` — provide field values, an MFA option ID, or an SSO button selector when the flow is `awaiting_input`. Agent inspects `discovered_fields` / `mfa_options` from `get` to know what's needed.
- **`manage_credentials`** (full CRUD)
  - `list` / `get` (SDK never returns values) / `totp_code` (current 6-digit code)
  - `create` / `update` / `delete` — agents can store and rotate credentials directly. Values, sso_provider, and totp_secret are all settable.
- **`manage_credential_providers`** (full CRUD)
  - `list` / `get` / `create` / `update` / `delete` for external providers (e.g. 1Password)
  - `list_items` to enumerate available items from the provider, and `test` to validate the token and list accessible vaults

## Agent Experience / Flow

This PR gives agents a durable path from "I need to use a logged-in site" to "I have a browser profile that can be reused for that site." The agent should first decide whether it already has a profile and credential source, then create the auth connection, start the login, and poll the connection state until the profile is healthy.

Typical flow:

1. Agent identifies the target domain and profile it wants to keep logged in.
2. Agent discovers or creates the credential source: `manage_credentials list/get/create`, or `manage_credential_providers list_items` when the secret lives in an external provider.
3. Agent creates the connection: `manage_auth_connections create domain=<site> profile_name=<profile> credential_name=<credential>`.
4. Agent starts the flow with `manage_auth_connections login id=<conn_id>` and shares the `hosted_url` with the user when human login is required.
5. Agent polls `manage_auth_connections get id=<conn_id>` until the flow reaches success, or until it sees `awaiting_input` with `discovered_fields` / `mfa_options`.
6. If MFA or additional fields are needed, agent collects the exact requested value, optionally calls `manage_credentials totp_code`, then uses `manage_auth_connections submit`.
7. Once the connection is healthy, future browser work starts with `manage_browsers create profile_name=<profile>` so the session opens already logged in.
8. Agent uses connection health/list/get before destructive actions and deletes only stale or explicitly requested connections/credentials.

Agent safety notes:

- `get` never returns credential secret values, so agents should not expect to recover passwords after creation.
- `totp_code` and provider tokens are sensitive outputs; agents should use them only for the active auth flow and avoid echoing them in summaries.
- Hosted login is still a user-facing boundary: the agent can orchestrate the flow, but the user may need to complete the site login.

## Test plan

- [ ] Local server starts; tools register without errors
- [ ] `manage_auth_connections list` returns existing connections for the authed user
- [ ] `manage_auth_connections create` + `login` returns a working hosted URL
- [ ] `manage_credentials` round-trip: `create` → `list` → `get` → `update` → `delete`
- [ ] `manage_credentials totp_code` returns a 6-digit code for a TOTP-enabled credential
- [ ] `manage_credential_providers` round-trip: `create` → `test` → `list_items` → `update` → `delete`

## Tool count

Bumps from 10 → 13.

---

> [!NOTE]
> **High Risk**
> Introduces tools that create/update/delete credentials, accept provider tokens, and return TOTP codes—high sensitivity even with OAuth-gated access.
> 
> **Overview**
> Adds **managed auth** to the MCP server so agents can keep browser profiles logged into third-party sites without ad-hoc API work.
> 
> Three new `manage_*` tools are registered in `register.ts` (and can be disabled via `KERNEL_MCP_DISABLED_TOOLSETS` like other toolsets): **`manage_auth_connections`** (create/list/get/delete, `login` with hosted + live view URLs, `submit` for MFA/SSO/fields, optional Kernel or external credentials and proxies), **`manage_credentials`** (CRUD plus `totp_code`; secrets are not returned on `get`), and **`manage_credential_providers`** (1Password-style providers with `list_items` and `test`). The README tool count is updated from 12 to **15**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 394ea3444414e2db58654bcab9e54ac6b092ffcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dcruzeneil2 (dismissed):
LGTM! Clean PR, follows existing codebase patterns well. One nit:

**nit:** `submit` in `manage_auth_connections` lets `fields: {}` through the validation guard since `!params.fields` is falsy for empty objects. The API will likely reject it anyway, but `manage_credentials create` already handles the equivalent case with `Object.keys(params.values).length === 0`. Worth adding the same check here for consistency.

masnwilliams (commented):
reviewed — three managed-auth tools (`manage_auth_connections`, `manage_credentials`, `manage_credential_providers`). all follow the `manage_*` pattern, typecheck, and every SDK call is verified against `@onkernel/sdk@0.58.0` (`auth.connections.*`, `credentials.*` incl. `totpCode`, `credentialProviders.*` incl. `listItems`/`test`). import + registration correct.

### code-quality positives
- `auth-connections.ts` create has the most thorough validation in any of these PRs — the mutual-exclusion across `credential_name` vs `credential_provider`+`credential_path` vs `credential_provider`+`credential_auto` is well done.
- good design: `credentials get` returns metadata only (values never returned), and the `login` flow hands back a `hosted_url` for the **user** to complete login rather than the agent handling raw passwords.

### consistency (only if this lands after #112)
these files predate the `responses.ts` helpers from #112 and use the verbose `{ content: [...] }` blocks. whichever of #104/#112 merges second should get a quick follow-up onto `textResponse`/`jsonResponse`/`errorMessage`. not blocking.

### agent-fit — the substantive bit
the read/list/login surface is well-designed for an agent. the concern is the **ingest** surface, where long-lived secrets arrive as plaintext tool arguments and therefore land in the conversation/tool-call transcript:
- `manage_credentials: create/update` — `values` (passwords) and `totp_secret` (a long-lived base32 2FA seed) as plaintext args.
- `manage_credential_providers: create/update` — `token` (e.g. a 1Password service-account token) as plaintext.
- `manage_credentials: totp_code` — emits a live 6-digit code; ephemeral (~30s), low concern, intended purpose.

for reference on how established servers handle this:
- **GitHub** keeps secrets **write-only** (Kernel already matches this on the read side — values never returned) and routes the token via **env var, not a tool arg**; also supports `--read-only` / `--toolsets` gating.
- **Stripe** exposes mutating tools but enforces **read-only scope server-side** via restricted keys.

none of the 10 MCP tools currently set risk annotations. suggested, in order of leverage:
1. add MCP annotations (`readOnlyHint` on list/get/totp, `destructiveHint` on delete) — drives client confirmation, zero behavior change.
2. toolset / read-only gating (GitHub model) so a deployment can opt out of credential management entirely.
3. confirm a read-only Kernel API key rejects `credentials.create`/`credentialProviders.create` server-side, so the guarantee isn't model-dependent.
4. the secret-as-tool-argument path has no clean industry precedent (GitHub deliberately routes secrets through env/restricted keys) — worth a deliberate decision rather than treating create/update as ordinary CRUD.

dcruzeneil2 (approved):
lgtm

PR #122, opened 2026-07-14
Update @onkernel/sdk to 0.78

yummybomb:
## summary

- update `@onkernel/sdk` from `^0.60.0` to `^0.78.0`
- migrate browser pool, extension, proxy, and credential-provider list calls to paginated SDK responses
- expose `limit`/`offset` and return `has_more`/`next_offset` for those list tools
- remove the unsupported browser-pool `save_profile_changes` input
- keep the fixed browser-pools resource exhaustive by iterating all pages

This isolates the SDK migration from feature PRs that need newer SDK endpoints. It supersedes the dependency migration in #117.

## verification

- `bun install --frozen-lockfile`
- `bunx tsc --noEmit`
- Prettier check on changed source and package files

The lockfile's `next` spec normalization (`16.2.6` to `^16.2.6`) matches the existing `package.json`; the resolved Next.js version is unchanged.

---

> [!NOTE]
> **Medium Risk**
> SDK upgrade plus changed list response shapes may break agents that assumed full arrays or the removed pool field; pagination behavior is a contract change across multiple tools.
> 
> **Overview**
> **Bumps `@onkernel/sdk` from `^0.60.0` to `^0.78.0`** (lockfile only normalizes `next` to `^16.2.6`; resolved Next version unchanged).
> 
> **List operations** on browser pools, extensions, proxies, and credential providers now call the SDK’s paginated `list({ limit, offset })` APIs and return results via `paginatedJsonResponse` with **`has_more` / `next_offset`**. Those tools accept shared **`limit` / `offset`** from `paginationParams`. `paginatedJsonResponse` forwards **`emptyText`** and other options into the uniform `{ items, has_more, next_offset }` JSON shape.
> 
> The **`browser-pools://` resource** still returns a full inventory by **async-iterating** all pages from `browserPools.list()`, while the **`list` tool** stays page-sized.
> 
> **Browser pool create/update** drops the **`save_profile_changes`** MCP input (omitted from `PoolConfigParams`) because the newer SDK/API no longer supports it.
> 
> **`manage_credentials`** tightens the **`values`** schema to `z.record(z.string(), z.string())` for Zod/SDK alignment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c02b14976c49d9b30c2d3cca95d91cbb35246814. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
         "@clerk/themes": "^2.4.19",
         "@mcp-ui/server": "^5.10.0",
         "@modelcontextprotocol/sdk": "1.26.0",
-        "@onkernel/sdk": "^0.60.0",
+        "@onkernel/sdk": "^0.78.0",
         "@types/jsonwebtoken": "^9.0.10",
         "@types/redis": "^4.0.11",
         "builtin-modules": "^5.0.0",
@@ -20,7 +20,7 @@
         "jszip": "^3.10.1",
         "lucide-react": "^0.534.0",
         "mcp-handler": "^1.1.0",
-        "next": "16.2.6",
+        "next": "^16.2.6",
         "next-themes": "^0.4.4",
         "playwright": "^1.49.1",
         "prettier": "^3.6.2",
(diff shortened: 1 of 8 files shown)


Sayan- (approved):
## Review: verifying the SDK 0.60 → 0.78 breaking-change migration

Approving. Notes on how this was verified, so a future reader can retrace it.

The concern with a bump this wide isn't whether the new code reads correctly — it's whether every breaking change was caught. Two breaks matter here: the `.list()` methods that returned bare arrays now return an offset-paginated `PagePromise`, and `save_profile_changes` was removed from the pool config. So the review targeted completeness.

**Checked against the installed 0.78 SDK types, not the PR description:**

- `browserPools.list`, `extensions.list`, `proxies.list`, and `credentialProviders.list` all now return `PagePromise<…OffsetPagination, T>`, and the code consumes them via the page shape (`getPaginatedItems()`, `has_more`, `next_offset`). Their `*ListParams` extend `OffsetPaginationParams`, so the `limit`/`offset` inputs are real.
- `save_profile_changes` is gone from the SDK entirely, so `Omit<BrowserCreateConfigParams, "save_profile_changes">` and dropping the MCP input is a genuine fix.

**Checked for missed call sites.** Every `.list()` in `src/` was reviewed. The other list tools (`apps`, `browsers`, `profiles`, `api-keys`, `projects`, `auth-connections`, `credentials`) were already on the paginated page pattern and aren't touched here — this PR migrates only the four that were still returning arrays.

**`tsc --noEmit` is the completeness check.** It passes clean. Old code returned arrays, the new SDK returns `PagePromise`, so any missed call site would be a type error. Zero errors means nothing stale survived.

**The `responses.ts` change is required by the migration.** `paginatedJsonResponse` now spreads `...options` so `emptyText` reaches `itemsJsonResponse`; without it, the empty-state note for `extensions`/`proxies` would be dropped when they moved off `itemsJsonResponse`.

**Remaining changes:**
- The `browser-pools://` resource switched to `for await (… of …list())`, so it stays exhaustive across pages instead of relying on a single array return.
- `credentials` `values` → `z.record(z.string(), z.string())` is the required two-arg form; the list call itself was already paginated.

**Verdict:** breaking changes handled completely. `tsc` clean, page shapes match the SDK, no missed call sites.

**One nit (non-blocking):** the description mentions a mocked credential-provider list test, but there's no test file in the diff and `bun test` finds none in the repo. The change is sound without it — either add the test or drop the line so the description matches what landed.

yummybomb:
> **One nit (non-blocking):** the description mentions a mocked credential-provider list test, but there's no test file in the diff and `bun test` finds none in the repo. The change is sound without it — either add the test or drop the line so the description matches what landed.

Dropped the line in the description about the mocked credential provider list test

PR #125, opened 2026-07-21
Fix execute_playwright_code snapshot tip

dprevoznik:
## Summary

The `execute_playwright_code` tool's `code` param tip recommended `await page._snapshotForAI()` for "a comprehensive page state snapshot." Two problems with that:

- It's a **whole-page dump** whose size scales with page complexity — a major context-window driver, which is the opposite of what we want by default.
- It's a Playwright internal that **isn't exposed in the stealth-patched execution runtime**, so it throws `TypeError: page._snapshotForAI is not a function`. Agents then fall back to an even larger `body.innerText()`.

## Change

Replaced the tip with a one-liner that lists the in-scope objects, makes the return contract explicit, and steers agents off full-page dumps toward the supported public API:

> Playwright/TypeScript code with `page`, `context`, and `browser` objects in scope; the value you `return` is sent back. Example: `await page.goto('https://example.com'); return await page.title();` Return only what you need — prefer a targeted selector (e.g. `await page.locator('h1').innerText()`) or a region-scoped snapshot (e.g. `await page.locator('main').ariaSnapshot()`) rather than dumping the whole page.

`page`/`context`/`browser` scope matches the [docs](https://docs.kernel.sh/browsers/playwright-execution). Description-only; no change to execution or output behavior. Typechecks clean.

## Also

Removed `tsconfig.tsbuildinfo` from tracking and added it to `.gitignore` — it's a generated, volatile TS build cache that shouldn't be versioned.

## Follow-ups (not in this PR)

The public docs (`kernel/docs → reference/mcp-server/tools/execute-playwright-code.mdx`) and the SDK example repos still reference `_snapshotForAI()` and need the same fix.

---

> [!NOTE]
> **Low Risk**
> Documentation and gitignore-only changes with no runtime or auth impact.
> 
> **Overview**
> Updates the **`execute_playwright_code`** MCP tool’s `code` parameter description so agents know `page`, `context`, and `browser` are in scope and that only the **`return`** value is echoed back. The old tip pointed at **`page._snapshotForAI()`**, which blows up context and isn’t available in the stealth runtime; the new copy steers toward small, targeted reads (`locator(...).innerText()`, region-scoped **`ariaSnapshot()`**) instead of whole-page dumps. **No execution or API behavior changes**—description only.
> 
> Also stops tracking **`tsconfig.tsbuildinfo`** and adds it to **`.gitignore`**, with a minor **`.gitignore`** newline fix for **`mcp-key.pem`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b58ed9550362ff6962d17e81d10d902c22bca50f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


masnwilliams (approved):
reviewed — clean. verified the description change against the executor and it's accurate.

- `src/lib/mcp/tools/playwright.ts:14` — new scope claim (`page`, `context`, `browser` all in scope + `return` value sent back) checks out against `kernel-images` `server/runtime/playwright-daemon.ts` (`new AsyncFunction('page','context','browser', code)`). steering away from `_snapshotForAI()` toward scoped extraction is the right call for token cost. `ariaSnapshot()` example is valid public API.
- `.gitignore` / `tsconfig.tsbuildinfo` — housekeeping looks right: ignore the incremental build cache + remove the tracked copy in the same PR.

lgtm 👍

PR #136, opened 2026-08-03
Report capability gaps agents hit, on our own wording

masnwilliams:
## Summary

Turns on PostHog's missing-capability signal: agents that can't find a tool for what they were asked to do now call `get_more_tools`, which records a `$mcp_missing_capability` event carrying a short description of the gap. Today an agent in that position either gives up silently or invents a tool name, and we see nothing.

## What changed

`src/lib/mcp/analytics.ts` only:

- `reportMissing: true` on the `instrument()` call, plus an explicit `missingCapabilityToolName` so the advertised name and the detected name come from one constant.
- We register the `get_more_tools` tool ourselves instead of letting `reportMissing` inject its descriptor (it only injects when no tool of that name exists). The reason is the `context` argument: the SDK's own descriptor asks for "a description of your goal and what kind of tool would help", and the SDK deliberately skips the hardened `context` description this server configures for every other tool. That field is the one intent value describing the user's *original ask*, so it's the likeliest to name a target site or an account — exactly what the rest of this integration is written to avoid capturing. Our version asks for the capability, not the specifics, and names what must never go in it.
- The tool description itself is kept as the SDK's wording, since that's the text that gets an agent to report a gap rather than give up. The callback returns the SDK's canned acknowledgement, and only runs if `instrument()` ever stops answering the call itself.

No new env var, no dependency change, and no allow-list change: `$mcp_intent` was already permitted, so the new event inherits the existing `beforeSend` — arguments are dropped and the intent is truncated at 300 chars.

## Verification

No test harness in this repo, so this was checked by hand against `bun run dev` with `POSTHOG_HOST` pointed at a local sink that decodes the capture payloads:

- `tools/list` returns exactly one `get_more_tools`, carrying our `context` description; every other tool keeps the existing per-call intent description.
- Calling it returns the acknowledgement and emits `$mcp_missing_capability` (not `$mcp_tool_call`), with `$mcp_intent`, `$mcp_intent_source: context_parameter`, and `$mcp_resource_name: get_more_tools`.
- An extra unexpected argument alongside `context` does not appear in the event — no `$mcp_parameters` survives `beforeSend`.
- A 500-char context is captured truncated to 300. A call with no `context` records an empty intent rather than erroring.

## Note

The tool is only advertised when `POSTHOG_PROJECT_TOKEN` is set, so a deployment without analytics doesn't get a tool that reports nowhere. It isn't wired into `KERNEL_MCP_DISABLED_TOOLSETS`, since it belongs to the analytics integration rather than a Kernel toolset.

The signal is only worth the tool slot if someone triages it — expect some reports where a tool does exist but was named or described badly, which is useful input of a different kind.

---

> [!NOTE]
> **Low Risk**
> Analytics-only changes in one file; no auth or data-path changes, and capability reporting is gated on PostHog being configured.
> 
> **Overview**
> Enables PostHog **missing-capability** tracking so agents can report tool gaps via **`get_more_tools`**, emitting **`$mcp_missing_capability`** with **`$mcp_intent`** instead of failing silently.
> 
> In **`instrumentMcpAnalytics`**, **`reportMissing: true`** and a shared **`missingCapabilityToolName`** tie into a **custom-registered** `get_more_tools` tool (SDK description kept; **`context`** uses hardened copy that avoids sites/accounts). **`beforeSend`** now runs **`sanitizeIntent`** (email/URL/token redaction plus 300-char cap), drops non-string intents, and **suppresses** missing-capability events with no usable context. The tool is only registered when **`POSTHOG_PROJECT_TOKEN`** is set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbdf3db19d10f5d64967115cdc0852d112ff105d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/mcp/analytics.ts
+++ b/src/lib/mcp/analytics.ts
@@ -67,6 +67,25 @@ const INTENT_ARGUMENT_DESCRIPTION =
 // stream a payload or a prompt into an event property.
 const INTENT_MAX_LENGTH = 300;
 
+// The intent descriptions ask agents to leave specifics out, and the agent writing the string
+// is the only thing holding them to it. These cover the shapes that are unambiguous when one
+// does slip through. They are not a substitute for the instruction: no pattern can tell that a
+// plain noun is a customer's name, so an intent still has to be treated as agent-written prose.
+const INTENT_REDACTIONS: readonly [RegExp, string][] = [
+  [/[^\s@]+@[^\s@]+\.[^\s@]+/g, "[email]"],
+  [/[a-z][a-z0-9+.-]*:\/\/\S+/gi, "[url]"],
+  [/\b[A-Za-z0-9_-]{32,}\b/g, "[token]"],
+];
+
+function sanitizeIntent(intent: string) {
+  const redacted = INTENT_REDACTIONS.reduce(
+    (text, [pattern, replacement]) => text.replace(pattern, replacement),
+    intent.trim(),
+  );
+
+  return redacted.slice(0, INTENT_MAX_LENGTH);
+}
+
 // Must stay the SDK's default name: reportMissing advertises a tool under this name and
 // dispatches calls to it as capability reports, and the name is what ties the two together.
 const MISSING_CAPABILITY_TOOL_NAME = "get_more_tools";
@@ -150,12 +169,25 @@ export function instrumentMcpAnalytics(server: McpServer) {
         if (!SENT_PROPERTIES.has(key)) delete properties[key];
       }
 
+      // Only a string is an intent. Anything else is a client sending an object or an array
+      // through the argument, which would land in PostHog as a serialized payload.
       const intent = properties[PostHogMCPAnalyticsProperty.Intent];
-      if (typeof intent === "string" && intent.length > INTENT_MAX_LENGTH) {
-        properties[PostHogMCPAnalyticsProperty.Intent] = intent.slice(
-          0,
-          INTENT_MAX_LENGTH,
-        );
+      const sanitized =
+        typeof intent === "string" ? sanitizeIntent(intent) : "";
+      if (sanitized) {
+        properties[PostHogMCPAnalyticsProperty.Intent] = sanitized;
+      } else {
+        delete properties[PostHogMCPAnalyticsProperty.Intent];
+      }
+
+      // The SDK answers a get_more_tools call before the registered schema validates it, so a
+      // report can arrive with no usable context: missing, blank, or not a string. The reported
+      // gap is the entire event, so drop the ones that don't carry one rather than count them.
+      if (
+        event.event === PostHogMCPAnalyticsEvent.MissingCapability &&
+        !sanitized
+      ) {
+        return null;
       }
 
       return event;
--- a/src/lib/mcp/analytics.ts
+++ b/src/lib/mcp/analytics.ts
@@ -1,5 +1,6 @@
 import {
   encodeSessionId,
+  getMoreToolsResult,
   instrument,
   MCP_SESSION_HEADER,
   newSessionId,
@@ -8,6 +9,7 @@ import {
 } from "@posthog/mcp";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { PostHog } from "posthog-node";
+import { z } from "zod";
 
 const projectToken = process.env.POSTHOG_PROJECT_TOKEN;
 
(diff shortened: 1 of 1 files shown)


yummybomb (approved):
nice implementation overall — reusing the SDK’s missing-capability flow while owning the descriptor keeps the behavior aligned and improves the privacy guidance.

### one observations before merge (if they are intended):

1. the privacy protection is advisory only. `context` remains free-form text, and the current sanitization does not redact URLs, domains, emails, account names, or most credentials embedded in that string. those values would still reach PostHog in `$mcp_intent`; the 300-character limit only truncates them.

2. `@posthog/mcp` intercepts `get_more_tools` before the registered Zod schema runs. calls with missing, blank, non-string, or extra arguments all succeed, and invalid/missing context produces an event without `$mcp_intent`. we should either enforce the contract in the interception path or drop invalid reports so the signal isn’t polluted.

masnwilliams:
Both were intended-as-shipped gaps rather than deliberate choices, and both are worth closing. Pushed a follow-up commit.

**2 (contract not enforced) — fixed.** Confirmed: the SDK matches `get_more_tools` in the call-tool handler and returns before the registered schema ever runs, so missing, blank, non-string, and extra arguments all reach capture. `beforeSend` now drops a `$mcp_missing_capability` event with no usable intent — a report whose entire content is the gap it describes says nothing without one. The agent still gets the same acknowledgement in every case; only the event is dropped. Extra arguments were already inert, since anything off the allow-list (including `$mcp_parameters`) is deleted before send.

Non-string context turned out to be the sharper half of this: intent was only truncated when `typeof intent === "string"`, so an object or array passed as `context` would have ridden through as a serialized payload on any tool, not just this one. Intent is now a string or it isn't sent.

**1 (advisory-only) — partly fixed, and the rest is inherent.** Agreed the description is guidance, not a control, and that's true of `$mcp_intent` on every tool, not just this one. Intent is now scrubbed for the shapes that are unambiguous when they do slip through — email addresses, `scheme://` URLs, and opaque 32+ char tokens — before truncation, on every event.

What deliberately isn't redacted is bare nouns: a company name, a product, a plain domain like `example.com`. No pattern distinguishes those from ordinary prose, and stripping them would remove the signal the event exists to collect ("agents want capability X for this class of site"). So the posture is: mechanical removal for secrets and personal identifiers, instruction plus a 300-char cap for everything else, and intents treated as agent-written prose downstream. If we want a stronger guarantee than that, the honest version is not capturing intent at all — worth deciding explicitly rather than leaning on the description.

Verified against a local dev server with the capture payloads decoded: no context, blank context, and an object context each produce **no** event (ack still returned); `"log into https://portal.example.com/account as ops@example.com with key sk_live_…"` is captured as `"log into [url] as [email] with key [token]"`; a normal intent passes through unchanged; and ordinary tool calls still record theirs.

masnwilliams:
Bugbot is right, and it caught the failure mode that mattered most here: a gap is usually *named* as a long `snake_case` or `kebab-case` identifier, so a plain 32-char length rule would have eaten the one substantive part of the report it exists to collect.

Fixed by dropping length as the signal. A credential is marked by either a vendor prefix (Kernel API keys are `sk_*`, plus `pk_`/`rk_`/`whsec_`/`ghp_`/`glpat_`/`github_pat_`/`xox*`) or an unbroken high-entropy run — mixed case containing a digit, or 32+ hex. Separator-joined lowercase names match none of those.

Checked both directions against the running server:

| intent | captured |
| --- | --- |
| `rotation for manage_credential_providers and create_browser_session_with_proxy_and_profile` | unchanged |
| `manage_api_keys style key_rotation_and_token_refresh support` | unchanged |
| `fan-out-over-browser-pool-with-shared-auth-state` | unchanged |
| `log into https://portal.example.com/account as ops@example.com with key sk_live_4eC39…` | `log into [url] as [email] with key [token]` |
| `Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOi…` | `Bearer [token].[token].[token]` |
| `profile id d41d8cd98f00b204e9800998ecf8427e` | `profile id [token]` |

One boundary worth naming rather than hiding: an opaque id carrying a non-vendor prefix (`sess_019fc86a…`) is left intact, because the hex run has no word boundary after the underscore. That's deliberate — resource ids aren't credentials, and they're often the useful half of a gap report. The rule targets credential shapes and personal identifiers, not identifiers in general.

PR #139, opened 2026-08-06
Align MCP browser pools with durable API contracts

IlyaasK:
## Summary

- align browser-pool validation with merged API limits and defaults
- add durable update clear semantics for profile, proxy, extensions, Chrome policy, and start URL
- return authoritative resolved profile and extension IDs in compact responses
- keep create and update semantics distinct for empty values

## Why

The MCP tool lagged the durable browser-pool API contract. Agents could not reliably pause filling, clear configuration, or see the IDs the API actually resolved.

## Implementation

The registered MCP surface now enforces integer fill rates, timeout bounds of 10 through 259200 seconds, and update-only empty sentinels. Tests exercise the registered tool rather than exporting implementation helpers.

## Verification

- `bun test` (90 tests)
- `bunx tsc --noEmit`
- Prettier and `git diff --check`
- registered-tool scenario tests for create, update, validation, omission, clears, and resolved IDs
- targeted Stryker runs killed all 46 core PATCH mutants; the expanded boundary run killed 45/51, with survivors limited to schema-blocked duplicate validation and description strings
- crap4ts: create 6.0, update 13.0, shared config 13.0

---

> [!NOTE]
> **Medium Risk**
> Changes MCP tool input mapping and API payloads for pool create/update (clears, bounds, responses); incorrect mapping could send wrong clears or reject valid agent requests, but scope is limited to browser-pool tooling with strong test coverage.
> 
> **Overview**
> Aligns the **`manage_browser_pools`** MCP tool with the durable browser-pool API so create/update behavior, validation, and responses match what agents need.
> 
> **Create vs update:** Empty `start_url` and `clear_profile` / `clear_extensions` are **update-only**; create rejects them. Updates can clear profile, extensions, proxy (`proxy_id: ""`), `start_url`, and Chrome policy (`{}`) via explicit sentinels, with guards against mixing clears with set profile/extension fields.
> 
> **Validation:** Integer `fill_rate_per_minute` (min 0) and `timeout_seconds` bounded **10–259200**; `start_url` schema allows `""` on update. Pool summaries expose **`profile_id`** and **`extension_ids`** from the API instead of nested profile/extension objects.
> 
> **Refactor:** **`buildBrowserSharedConfig`** extracts shared profile/extensions/viewport building; create still uses **`buildBrowserCreateConfig`** for `start_url`. Kernel client test mocks move to **`kernel-client.test-fixtures`**; new contract tests cover the registered tool.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08f762c287113a975323f948101b04c00713b405. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/mcp/tools/browser-pools.test.ts
+++ b/src/lib/mcp/tools/browser-pools.test.ts
@@ -1,7 +1,7 @@
 /// <reference types="bun-types" />
 
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import { beforeEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import type { z } from "zod";
 import {
   kernelClientMock,
@@ -60,6 +60,10 @@ function pool() {
 
 describe("browser-pool contract parity", () => {
   beforeEach(() => {
+    resetKernelClientFactory();
+  });
+
+  afterEach(() => {
     resetKernelClientFactory();
   });
 
--- a/src/lib/mcp/tools/browser-pools.test.ts
+++ b/src/lib/mcp/tools/browser-pools.test.ts
@@ -1,0 +1,240 @@
+/// <reference types="bun-types" />
+
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { beforeEach, describe, expect, mock, test } from "bun:test";
+import type { z } from "zod";
+
+let kernelClient: unknown;
+
+mock.module("@/lib/mcp/kernel-client", () => ({
+  createKernelClient: () => kernelClient,
+}));
+
+const { registerBrowserPoolCapabilities } = await import(
+  "@/lib/mcp/tools/browser-pools"
+);
+
+type ToolHandler = (
+  params: Record<string, unknown>,
+  extra: { authInfo?: { token: string } },
+) => Promise<{
+  content: Array<{ type: string; text: string }>;
+  isError?: boolean;
+}>;
+
+function captureBrowserPoolTool() {
+  let handler: ToolHandler | undefined;
+  let schema: z.ZodRawShape | undefined;
+  const server = {
+    resource() {},
+    registerResource() {},
+    tool(name: string, ...args: unknown[]) {
+      if (name !== "manage_browser_pools") return;
+      schema = args[1] as z.ZodRawShape;
+      handler = args.at(-1) as ToolHandler;
+    },
+  } as unknown as McpServer;
+
+  registerBrowserPoolCapabilities(server);
+  if (!handler || !schema)
+    throw new Error("manage_browser_pools was not registered");
+  return { handler, schema };
+}
+
+const auth = { authInfo: { token: "test-token" } };
+
+function pool() {
+  return {
+    id: "pool_1",
+    name: "production",
+    created_at: "2026-08-06T12:00:00Z",
+    available_count: 2,
+    acquired_count: 1,
+    profile_id: "profile_resolved",
+    extension_ids: ["extension_resolved"],
+    browser_pool_config: {
+      size: 3,
+      profile: { name: "configured-profile-name" },
+      extensions: [{ name: "configured-extension-name" }],
+    },
+  };
+}
+
+describe("browser-pool contract parity", () => {
+  beforeEach(() => {
+    kernelClient = undefined;
+  });
+
+  test("validates API boundaries", () => {
+    const { schema } = captureBrowserPoolTool();
+
+    expect(schema.fill_rate_per_minute.safeParse(0).success).toBe(true);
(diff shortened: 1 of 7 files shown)


src/lib/mcp/tools/browser-pools.ts
```diff
@@ -42,8 +42,13 @@ type PoolConfigParams = Omit<
   fill_rate_per_minute?: number;
   chrome_policy?: Record<string, unknown>;
   kiosk_mode?: boolean;
+  clear_profile?: boolean;
+  clear_extensions?: boolean;
 };
 
+const browserPoolTimeoutSchema = z.number().int().min(10).max(259200);
```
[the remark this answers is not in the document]
IlyaasK: Verified against merged kernel/kernel origin/main: BrowserPoolConfig, BrowserPoolCreateRequest, and BrowserPoolUpdateRequest all declare fill_rate_per_minute as type integer. The handler receives an int, validates it against the organization-specific integer cap, stores it as int32, and returns an int. Fractional values are therefore not part of the API contract, so the MCP .int() constraint is intentional and this finding is not actionable.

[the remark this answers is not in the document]
IlyaasK: experimentally invalidated

src/lib/mcp/tools/browser-pools.ts
```diff
@@ -94,10 +111,34 @@ function buildPoolUpdateParams(
...
+      error:
+        "Error: clear_extensions cannot be combined with extension_id or extension_name.",
+    };
+  }
+
   return {
     ok: true,
     value: {
       ...config.value,
+      ...(params.proxy_id !== undefined && { proxy_id: params.proxy_id }),
```
masnwilliams, line 141: this repair step shows that PATCH clear semantics are being routed through a create-shaped serializer: empty `proxy_id` and `chrome_policy` are dropped in `buildPoolConfigParams` and re-added here, while empty `start_url` is supported by weakening the shared `buildBrowserCreateConfig` helper. that helper now accepts `buildBrowserCreateConfig({ start_url: "" })`, even though an empty URL is update-only, so browser creation stays safe only because its outer Zod schema happens to reject it first.

can we keep the shared create helper strict and separate the genuinely common profile/extension/viewport translation from operation-specific fields? then create can normalize its inputs once, while update can preserve its PATCH sentinels directly. that removes the drop-and-restore flow and keeps clear behavior at the update boundary.

src/lib/mcp/tools/browser-pools.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    browser_pool_config: {
+      size: 3,
+      profile: { name: "configured-profile-name" },
+      extensions: [{ name: "configured-extension-name" }],
+    },
+  };
+}
+
+describe("browser-pool contract parity", () => {
+  beforeEach(() => {
```
masnwilliams, line 62: the centralized factory is process-global, but this suite resets it only before each test. the final test leaves `kernelClientMock.factory = () => ({ browserPools: {} })` installed, and Bun shares that module state with later test files. i reproduced this by importing this suite and checking the default factory afterward; it returned the leftover browser-pool stub instead of throwing `unexpected Kernel client use`.

please restore it in `afterEach` as well, or use a scoped helper that installs a fake for one test and always restores it. otherwise the change intended to make mocks order-independent still leaks state across suites.

masnwilliams (changes requested):
the contract fixes are needed and the behavior looks correct, but i think these two maintainability issues should be addressed before merging. the main concern is keeping PATCH-only clear semantics out of the shared create path; the second is ensuring the new shared test mock is actually isolated between suites.

src/lib/mcp/tools/browser-pools.ts
IlyaasK: Fixed in 08f762c. `buildBrowserCreateConfig` is strict again and rejects an empty `start_url`. The new shared builder translates only profile, extensions, and viewport. `buildPoolUpdateParams` now handles PATCH-specific `start_url`, `proxy_id`, and `chrome_policy` values directly, so clear sentinels are preserved without a drop-and-restore path. I also added a direct regression test for the create helper.

src/lib/mcp/tools/browser-pools.test.ts
IlyaasK: Fixed in 08f762c. The browser-pool suite now restores the default Kernel client factory in `afterEach`, so the final test cannot leak its stub into a later suite. The full suite passes: 91 tests across 17 files.

masnwilliams (approved):
both requested changes are addressed: create serialization is strict again with PATCH clears owned by the update path, and the shared Kernel client factory is restored after the browser-pool suite. re-ran the full test suite, typecheck, changed-file formatting, and the cross-suite leak reproduction; all pass.

PR #140, opened 2026-08-06
Add durable profile and proxy MCP contract parity

IlyaasK:
## Summary

- make profile setup use exact-name filtering and reject ambiguity
- add profile and proxy rename actions
- preserve stable API error codes in MCP failures
- bind profile setup browsers to the resolved profile ID

## Why

Durable MCP operations should resolve stable identities and surface actionable API failures. Search-style profile matching and name reuse could otherwise select the wrong durable profile.

## Implementation

Profile setup filters by exact name, rejects duplicate exact matches with an executable recovery path, and uses the resolved ID when opening the setup browser. Rename guidance warns before invocation about active profile sessions. Rename responses retain their direct SDK resource shape.

## Verification

- `bun test` (95 tests)
- `bunx tsc --noEmit`
- Prettier and `git diff --check`
- 20 focused scenario tests covering exact lookup, ambiguity, missing resources, rename guards, stable profile identity, and coded API errors
- targeted Stryker: profile/proxy rename and exact-lookup core paths reached 100%; expanded setup/error run killed 84/106 mutants, with remaining survivors in pre-existing presentation branches and equivalent API-error guards
- crap4ts: changed registration paths 1.1; API error-code helper 5.0

---

> [!NOTE]
> **Medium Risk**
> Changes how durable profiles are resolved during setup (ID binding and ambiguity handling), which could alter agent behavior if duplicate names existed; production wiring stays on default dependencies.
> 
> **Overview**
> Introduces **`McpDependencies`** so MCP registration and tools can inject `createKernelClient` (defaults unchanged in production), enabling contract tests without hitting the real API.
> 
> **Profile setup** now lists by exact `name`, fails on duplicate exact matches with recovery guidance, rejects `update_existing` when the profile is missing, and opens the setup browser with **`profile.id`** instead of name. **`manage_profiles`** gains a **`rename`** action (by name or ID) with shared identifier validation; tool docs warn against renaming while a browser session is attached.
> 
> **`manage_proxies`** gains **`rename`** via `proxies.update`. MCP **`throwToolError`** messages append **`[code: …]`** when the Kernel SDK returns a string API error code.
> 
> Adds **`durable-contracts.test.ts`** and extends **`responses.test.ts`** for coded errors and rename/schema behavior at the MCP boundary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7fa9aaa19e346fd276dcf0b6f65e92cd3d4d217. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/mcp/kernel-client.test-fixtures.ts
+++ b/src/lib/mcp/kernel-client.test-fixtures.ts
@@ -10,7 +10,7 @@ export const unusedKernelClient = new Proxy(
 );
 
 export const kernelClientMock: {
-  factory: (token: string) => any;
+  factory: (token: string) => unknown;
 } = {
   factory: () => unusedKernelClient,
 };
--- a/src/lib/mcp/kernel-client.test-fixtures.ts
+++ b/src/lib/mcp/kernel-client.test-fixtures.ts
@@ -10,7 +10,7 @@ export const unusedKernelClient = new Proxy(
 );
 
 export const kernelClientMock: {
-  factory: (token: string) => unknown;
+  factory: (token: string) => any;
 } = {
   factory: () => unusedKernelClient,
 };
--- a/src/lib/mcp/tools/durable-contracts.test.ts
+++ b/src/lib/mcp/tools/durable-contracts.test.ts
@@ -1,16 +1,13 @@
 /// <reference types="bun-types" />
 
-import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import { afterEach, describe, expect, test } from "bun:test";
-import {
-  kernelClientMock,
-  resetKernelClientFactory,
-} from "@/lib/mcp/kernel-client.test-fixtures";
-
-const { registerProfileCapabilities } = await import(
-  "@/lib/mcp/tools/profiles"
-);
-const { registerProxyTools } = await import("@/lib/mcp/tools/proxies");
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { describe, expect, test } from "bun:test";
+import type { McpDependencies } from "@/lib/mcp/dependencies";
+import type { KernelClient } from "@/lib/mcp/kernel-client";
+import { registerProfileCapabilities } from "@/lib/mcp/tools/profiles";
+import { registerProxyTools } from "@/lib/mcp/tools/proxies";
 
 type ToolResult = {
   content: Array<{ type: string; text: string }>;
@@ -22,7 +19,15 @@ type ToolHandler = (
   extra: { authInfo?: { token: string } },
 ) => Promise<ToolResult>;
 
-function captureTool(register: (server: McpServer) => void, name: string) {
+type RegisterTool = (server: McpServer, dependencies?: McpDependencies) => void;
+
+function testDependencies(client: unknown): McpDependencies {
+  return {
+    createKernelClient: () => client as KernelClient,
+  };
+}
+
+function captureTool(register: RegisterTool, name: string, client: unknown) {
   let handler: ToolHandler | undefined;
   const server = {
     resource() {},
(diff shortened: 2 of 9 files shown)


src/lib/mcp/tools/profiles.ts
```diff
@@ -110,13 +115,20 @@ export function registerProfileCapabilities(server: McpServer) {
...
-            const existingProfile = existingProfiles?.find(
-              (p) => p.name === params.profile_name,
-            );
+            const existingProfiles = await listProfiles(client, {
+              name: params.profile_name,
+            });
+            if (existingProfiles.length > 1) {
+              return errorResponse(
+                `Error: multiple profiles match the exact name "${params.profile_name}". Rename or delete duplicate profiles by ID, then retry setup.`,
+              );
```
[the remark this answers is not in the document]
IlyaasK: Fixed in 89ce449. The ambiguity response now lists every exact match as `name (ID: ...)`, so the agent has the identifiers required to rename or delete duplicates. The contract test asserts both IDs are present.

src/lib/mcp/kernel-client.test-fixtures.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+);
+
+export const kernelClientMock: {
+  factory: (token: string) => unknown;
+} = {
+  factory: () => unusedKernelClient,
+};
+
+mock.module("@/lib/mcp/kernel-client", () => ({
```
masnwilliams: This replaces local duplication with process-global mutable test state. Importing this fixture permanently redirects `kernel-client` for the test process, every suite shares one mutable factory, and each consumer has to remember its own reset protocol. The top-level dynamic imports in `durable-contracts.test.ts` are evidence of that hidden ordering requirement; this will be unsafe under concurrent tests and brittle under ordinary import refactors. Can we inject the client factory into tool registration through a shared dependency object instead, so each test owns an isolated fake and the module graph stays side-effect free?

src/lib/mcp/tools/durable-contracts.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  extra: { authInfo?: { token: string } },
+) => Promise<ToolResult>;
+
+function captureTool(register: (server: McpServer) => void, name: string) {
+  let handler: ToolHandler | undefined;
+  const server = {
+    resource() {},
+    tool(toolName: string, ...args: unknown[]) {
+      if (toolName === name) {
+        handler = args.at(-1) as ToolHandler;
```
masnwilliams, line 36: These are named contract tests, but this harness discards the registered schema, casts out of `McpServer`, and invokes the callback directly with unvalidated objects. That verifies internal handler behavior, not that an MCP client can discover or call the new rename contract; regressions in `tools/list`, Zod validation, or MCP result serialization can pass. Please use the real `McpServer` + `Client` + `InMemoryTransport` pattern already present in `responses.test.ts`, backed by the injected fake client. Focused handler tests can remain, but profile and proxy rename each need boundary coverage.

masnwilliams (changes requested):
The production behavior is correct and worth landing: exact profile resolution, ID-bound setup sessions, rename parity, and stable API error codes all close real gaps. I'm requesting changes for the test architecture, which introduces global/import-order coupling and does not exercise the MCP contract it claims to cover. Once the tests use an isolated client dependency and validate the real MCP boundary, this should be good to merge. As a smaller follow-up while touching this code, please also consider centralizing the repeated profile_name/profile_id XOR validation used by get, rename, and delete.

src/lib/mcp/kernel-client.test-fixtures.ts
IlyaasK: Fixed in c7fa9aa. Profile and proxy registration now accept a shared `McpDependencies` object with a Kernel client factory; production uses the existing default factory, while each durable-contract test owns an isolated fake. The profile collection and resource-template paths receive the same dependency. `durable-contracts.test.ts` now uses static imports and no longer imports or mutates the process-global test fixture.

src/lib/mcp/tools/durable-contracts.test.ts
IlyaasK: Fixed in c7fa9aa. Profile and proxy rename now each have boundary coverage using a real `McpServer`, `Client`, and linked `InMemoryTransport`. The tests discover the tool through `tools/list`, prove invalid field types are rejected before the fake SDK runs, pass authenticated requests through the injected factory, execute valid rename calls, and assert the serialized MCP result. Focused handler tests remain for setup and missing/conflicting-field behavior.

IlyaasK:
Also completed the optional cleanup from the review summary in c7fa9aa: `get`, `rename`, and `delete` now share one profile identifier validator while preserving their existing action-specific diagnostics. Focused regressions cover both conflicting and missing identifiers for get and delete. Full suite: 109 tests passed; typecheck, managed-auth bundle check, and changed-file formatting passed.

IlyaasK:
> **Additional Suggestion:**
> 
> `fill_rate_per_minute` schema gained an `.int()` constraint that wrongly rejects the fractional percentage values the Kernel API accepts.
> 
> [Fix on Vercel](https://vercel.com/onkernel/~/agent/tasks/1af15344-67c0-4bde-8fdf-3946a562aa5e)

experimentally invalidated

masnwilliams (approved):
Rechecked c7fa9aa. Both requested structural changes are addressed: profile/proxy registrations now receive an isolated Kernel client dependency, and rename behavior is covered through a real MCP server/client boundary including discovery, validation, authentication propagation, SDK calls, and serialization. The repeated profile selector validation is also centralized. Local verification passes: 109 tests, TypeScript, changed-file Prettier, and diff checks. Approved.

PR #147, opened 2026-08-07
Check durable MCP top-level parameter-name parity

IlyaasK:
## Summary

- compare the registered top-level input parameter names for `manage_browser_pools`, `manage_profiles`, and `manage_proxies` with their public documentation tables
- report both schema parameters missing from docs and documented parameters missing from the schema
- deliberately follow `kernel/docs` `main` so changes in either repository can reveal cross-repository drift
- sparse-check out only the three relevant public docs pages in CI

## Scope

This checks **top-level parameter-name parity**, not complete schema equivalence. It does not compare types, requiredness, descriptions, enum values, or nested fields.

The durable tools covered by the gate are listed explicitly in `documentedTools`. Newly documented durable tools must be added there with their documentation path.

## Why CI follows docs main

The docs checkout intentionally has no pinned `ref`. Pinning it would make CI reproducible for one docs revision, but it would also prevent a later docs-only change from exposing drift against the MCP server. Tracking docs `main` makes the gate useful from either side of the cross-repository contract.

## Dependency

kernel/docs#489 has merged and supplies the matching parameter tables.

## Verification

- `bun test` — 114 passing
- `bun run check:managed-auth-app`
- focused parity tests — 5 passing
- before kernel/docs#489 merged, the checker failed with the expected missing-parameter diagnostics
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
         run: bun test
 
       - name: Check out Kernel docs
+        # Track docs main deliberately so either repository can detect cross-repo drift.
         uses: actions/checkout@v4
         with:
           repository: kernel/docs
@@ -40,5 +41,5 @@ jobs:
             reference/mcp-server/tools/manage-proxies.mdx
           sparse-checkout-cone-mode: false
 
-      - name: Check MCP documentation schema parity
-        run: bun scripts/check-doc-schema-parity.ts "$GITHUB_WORKSPACE/kernel-docs"
+      - name: Check MCP documentation top-level parameter-name parity
+        run: bun scripts/check-doc-top-level-parameter-name-parity.ts "$GITHUB_WORKSPACE/kernel-docs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,17 @@ jobs:
 
       - name: Test
         run: bun test
+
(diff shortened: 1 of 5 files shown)


masnwilliams (changes requested):
requesting changes on two structural issues:

### 1. required CI depends on mutable docs `main`

this makes an unchanged PR pass or fail based on whatever happens to be on `kernel/docs` main when CI runs. it also doesn't detect docs-side drift immediately, since changes in the docs repo don't trigger this workflow. can we keep PR CI deterministic by pinning a docs revision or contract snapshot, then run the main-to-main comparison separately as a scheduled or repository-dispatch drift check?

this is already affecting the PR: its current merge result fails because newer MCP schemas include `project_id`, while the docs do not.

### 2. coverage is duplicated across several manual lists

adding another covered tool currently requires updating the imports, `documentedTools`, explicit registration calls, and the workflow's sparse-checkout list. forgetting any one of those silently leaves the tool unchecked.

can we check out the tools directory, discover the relevant `manage-*.mdx` files, and list schemas through the canonical MCP registration path? if only these three tools should be covered, that subset should still be defined once rather than repeated across TypeScript and YAML.

non-blocking follow-up: please add an integration test around the complete checker using local fixture docs. the current tests cover Markdown parsing and set comparison, but not registration, document loading, missing files, or error aggregation.

PR #151, opened 2026-08-11
Attribute MCP analytics events to Kernel organizations

rgarcia:
## What

Every `$mcp_tool_call` / `$mcp_tools_list` / `$mcp_initialize` event now carries organization attribution via `$groups = { organization: org_id }` — the same convention the Kernel API's `api_call` events use. Today only the initialize event gets `$groups` (via `enrichMcpAnalyticsEvent`); tool-call and tools/list events have no org attribution at all.

Attribution goes through a single mechanism: the `eventProperties` hook stamps `$groups` on every captured event, reading `connectionContext.scope.organizationId` out of the request extras. The route already resolves the connection context at auth time on every request (`handleMcpRequestWithIdentity`), so this adds **zero I/O**. `identify` stays `null` and distinct-id semantics are unchanged (session-scoped `ses_` everywhere; the OAuth-initialize carve-out in `enrichMcpAnalyticsEvent` still uses the canonical Kernel user id).

Why `eventProperties` rather than the SDK's `identify` callback: the pinned SDK only runs `identify` on initialize and tools/call — never for tools/list — and mcp-handler builds a fresh `McpServer` per HTTP request, so the SDK's per-session identity cache is always cold when a tools/list request arrives (verified empirically against the pinned SDK). `eventProperties` runs on all three event types.

## Privacy posture (unchanged)

- `$process_person_profile` behavior is exactly main's: the SDK sets it false on every event (no identity is ever resolved), and `enrichMcpAnalyticsEvent` deletes it only for the OAuth-initialize carve-out, as before.
- `SENT_PROPERTIES` stays an allow-list — no new properties added, `$set` stays excluded, call payloads (`$mcp_parameters` / `$mcp_response` / `$mcp_error_message`) stay out.

## Failure behavior

When `authInfo.extra.connectionContext` is absent, the event goes out without `$groups` — main's current behavior. The read is a pure property access on request extras, so no latency is added to tool calls.

## Tests

- `sanitizeMcpAnalyticsEvent` (extracted from the inline `beforeSend`, unchanged behavior): `$groups` survives the allow-list, payloads/`$set` stripped, intent redaction, missing-capability/exception dropping.
- SDK-integration test (`instrumentMcpAnalytics` with a stub PostHog client, fresh `McpServer` per simulated request — the deployed topology): asserts initialize, tools/list, and tools/call all carry `$groups`, that distinct ids stay session-scoped, that no `$identify` is ever sent, and that events stay group-less without a connection context. Mutation-checked: fails if `$groups` stamping is removed.

`bun test`: 185 pass. `tsc --noEmit`, production build, and prettier clean. (`make lint` is broken pre-existing.)

🤖 Iterated with automated fable-5 reviews: first caught the tools/list attribution gap (fixed via eventProperties), then simplified to the single-mechanism design per review feedback.


src/lib/mcp/analytics.ts
```diff
@@ -221,56 +273,44 @@ export function instrumentMcpAnalytics(server: McpServer) {
...
+    // Attributes every event to the caller's organization via $groups — the same
+    // convention as the Kernel API's own events (api_call sends $groups with
+    // organization = org id). The distinct id the SDK requires in order to carry
+    // groups is an org-pseudonymous hash, never a person id, and beforeSend pins
+    // $process_person_profile to false, so no person profiles are created from it.
+    // The identity comes from the connection context the route resolves at auth
+    // time, so this adds no I/O per request. Events stay session-scoped beyond
+    // that: the initialize event alone uses the canonical Kernel user ID when auth
+    // context identifies a user (see enrichMcpAnalyticsEvent).
+    identify: (_request, extra) => resolveMcpOrgIdentity(extra),
```
masnwilliams: this `identify` hook is unnecessary now that `eventProperties` below stamps `$groups` on every event. keeping both mechanisms changes `distinct_id` to three different identity spaces in one MCP session (canonical user for OAuth initialize, `ses_*` for tools/list, and `mcporg_*` for tool calls), then requires the new hash helper, `$identify` suppression, and `$process_person_profile` compensation. none of that is needed for organization attribution. can we leave `identify: null`, read `connectionContext.scope.organizationId` directly in `eventProperties`, and preserve the existing distinct ids? that deletes the entire new identity layer while still putting `$groups.organization` on initialize, tools/list, and tools/call.

masnwilliams (changes requested):
the rebase removed the duplicate I/O and cache path, but the implementation still uses two attribution mechanisms where one is sufficient. `eventProperties` already covers initialize, tools/list, and tools/call, so the `identify` path adds a second identity model without being needed for `$groups`. please keep the existing distinct-id semantics and stamp the organization group from the route-resolved connection context through `eventProperties` alone. that should also let us delete `org-context.ts`, the synthetic hash, `$identify` suppression, and most of the person-profile compensation logic.

validation: 192 tests pass and `tsc --noEmit` passes after installing the locked dependencies.

src/lib/mcp/analytics.ts
rgarcia: Done in fb6d17c — agreed, one mechanism is better. `identify` is back to `null`, distinct ids keep their existing semantics (`ses_*` everywhere except the OAuth-initialize carve-out), and `eventProperties` now reads `connectionContext.scope.organizationId` directly from the request extras. Deleted org-context.ts + its test, the mcporg_ hash, the $identify drop, and the $process_person_profile pin — the sanitizer is back to main's behavior (enrich + allow-list + intent), just extracted so the tests can drive it. The SDK-integration test now asserts all three event types carry $groups with session-scoped distinct ids. 185 tests pass, tsc/build/prettier clean.

rgarcia:
Addressed in fb6d17c: single attribution mechanism via `eventProperties` only. `identify: null` restored, existing distinct-id semantics preserved, and the whole identify layer is gone (org-context.ts, the synthetic hash, $identify suppression, $process_person_profile compensation). `eventProperties` reads `connectionContext.scope.organizationId` from the request extras — resolved by the route at auth time on every request, so still zero added I/O. PR description rewritten to match.

masnwilliams (approved):
latest commit addresses the blocking feedback: organization groups now flow through eventProperties alone, existing distinct-id semantics are preserved, and the synthetic identity layer is gone. verified locally: 185 tests pass, tsc passes, and the changed files pass Prettier. CI, Semgrep, and BugBot are green.

PR #154, opened 2026-08-12
Allow raw telemetry page reads

rgarcia:
## summary

- add opt-in raw telemetry reads while preserving compact output by default
- enforce raw reads with explicit categories, `limit <= 5`, and a 1 MiB serialized-response cap
- reject screenshot PNGs from raw JSON responses
- preserve API-owned time and tail semantics; label replay arguments as best-effort because archives can change
- document omitted fields, output bounds, and replay consistency

## verification

- `bun test` (198 tests)
- `bunx tsc --noEmit`
- `bunx prettier --check src/lib/mcp/tools/browsers.ts src/lib/mcp/tools/browsers.test.ts`

---

> [!NOTE]
> **Medium Risk**
> Changes MCP telemetry response shape and size limits for agents debugging sessions; raw mode can expose sensitive network/console payloads but is gated and capped.
> 
> **Overview**
> Adds **opt-in raw telemetry** for `manage_browsers` `get_telemetry`: default compact output is unchanged, but callers can set `compact=false` with explicit categories and a **limit of 1–5** to get full upstream event envelopes (headers, bodies, etc.) instead of stripped fields.
> 
> Compact, category-filtered pages now include **`raw_replay_best_effort`**—tool arguments to re-request the same cursor page in raw mode (best-effort; retention/timing may differ). Raw mode **rejects screenshot category**, blocks **`png` in event data**, and **errors if the serialized page exceeds 1 MiB**.
> 
> `registerBrowserCapabilities` accepts injectable **`McpDependencies`** (kernel client factory) so browser tools/resources are testable; new **`browsers.test.ts`** covers replay, time-window semantics, limits, and resource reads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53ca6b7d3e2a69a95c45d55b4ce5347cec19bc9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/mcp/tools/browsers.ts
+++ b/src/lib/mcp/tools/browsers.ts
@@ -329,27 +329,36 @@ export function registerBrowserCapabilities(
   server: McpServer,
   dependencies: McpDependencies = defaultMcpDependencies,
 ) {
-  registerJsonResourceCollection(server, {
-    name: "browsers",
-    uriTemplate: "kernel://orgs/{organizationId}/projects/{projectId}/browsers",
-    emptyText: "No browsers found",
-    read: async (client) => {
-      const browsers = [];
-      for await (const browser of client.browsers.list()) {
-        browsers.push(browser);
-      }
-      return browsers;
+  registerJsonResourceCollection(
+    server,
+    {
+      name: "browsers",
+      uriTemplate:
+        "kernel://orgs/{organizationId}/projects/{projectId}/browsers",
+      emptyText: "No browsers found",
+      read: async (client) => {
+        const browsers = [];
+        for await (const browser of client.browsers.list()) {
+          browsers.push(browser);
+        }
+        return browsers;
+      },
     },
-  });
+    dependencies,
+  );
 
-  registerJsonResourceTemplate(server, {
-    name: "browser",
-    uriTemplate:
-      "kernel://orgs/{organizationId}/projects/{projectId}/browsers/{sessionId}",
-    variableName: "sessionId",
-    resourceLabel: "Browser session",
-    read: (client, sessionId) => client.browsers.retrieve(sessionId),
-  });
+  registerJsonResourceTemplate(
+    server,
+    {
+      name: "browser",
+      uriTemplate:
+        "kernel://orgs/{organizationId}/projects/{projectId}/browsers/{sessionId}",
+      variableName: "sessionId",
+      resourceLabel: "Browser session",
+      read: (client, sessionId) => client.browsers.retrieve(sessionId),
+    },
+    dependencies,
+  );
 
   // manage_browsers -- Manage browser sessions and read archived telemetry
   server.tool(
--- a/src/lib/mcp/tools/browsers.ts
+++ b/src/lib/mcp/tools/browsers.ts
@@ -6,7 +6,11 @@ import {
   buildBrowserUpdateConfig,
   type BrowserConfigResult,
 } from "@/lib/mcp/browser-config";
-import { createKernelClient, type KernelClient } from "@/lib/mcp/kernel-client";
+import {
+  defaultMcpDependencies,
+  type McpDependencies,
+} from "@/lib/mcp/dependencies";
+import type { KernelClient } from "@/lib/mcp/kernel-client";
 import {
   registerJsonResourceCollection,
   registerJsonResourceTemplate,
@@ -201,6 +205,7 @@ type BrowserTelemetryReadParams = {
   since?: string;
   until?: string;
   order?: "asc" | "desc";
+  compact?: boolean;
 };
 
 async function readBrowserTelemetry(
@@ -233,7 +238,9 @@ async function readBrowserTelemetry(
   const fullSessionRead = unfilteredExceptSince && params.since === undefined;
 
   const page = await client.browsers.telemetry.events(params.session_id, query);
-  const items = page.getPaginatedItems().map(compactTelemetryEvent);
+  const pageItems = page.getPaginatedItems();
+  const items =
+    params.compact === false ? pageItems : pageItems.map(compactTelemetryEvent);
 
   // Counter-steer the pagination reflex: an unfiltered ascending read starts
   // at session creation, so an agent chasing a recent failure should flip to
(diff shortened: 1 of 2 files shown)


src/lib/mcp/tools/browsers.ts
```diff
@@ -344,7 +354,7 @@ export function registerBrowserCapabilities(server: McpServer) {
   // manage_browsers -- Manage browser sessions and read archived telemetry
   server.tool(
     "manage_browsers",
-    'Manage browser sessions and their archived telemetry. Use "list" to choose an existing session, "create" before browser control, "update" to change supported session settings, "get" for full details, "get_telemetry" to diagnose active or deleted sessions, and "delete" when finished.',
```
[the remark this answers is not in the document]
rgarcia: Fixed in 61eb4e6 by passing the injected dependencies to both browser resource registrations. The MCP test now reads both browser resources through the injected client as well.

src/lib/mcp/tools/browsers.ts
```diff
@@ -233,7 +278,9 @@ async function readBrowserTelemetry(
   const fullSessionRead = unfilteredExceptSince && params.since === undefined;
 
   const page = await client.browsers.telemetry.events(params.session_id, query);
-  const items = page.getPaginatedItems().map(compactTelemetryEvent);
+  const pageItems = page.getPaginatedItems();
+  const items =
+    params.compact === false ? pageItems : pageItems.map(compactTelemetryEvent);
```
masnwilliams, line 277: this removes every output bound in raw mode while retaining the 100-event default. screenshot telemetry can carry roughly 1 MB of base64 PNG per event, and network events can include large bodies and headers, so one ordinary `compact: false` call can serialize tens to roughly 100 MB into a text response. schema prose asking the model to narrow the query isn’t an enforceable boundary. please require an explicit small raw limit, cap total serialized bytes, and keep `png` out of JSON (or expose it through a dedicated image/event path). add a regression for the oversized/default-limit case.

src/lib/mcp/tools/browsers.ts
```diff
@@ -195,33 +199,74 @@ function compactTelemetryEvent({ seq, event }: TelemetryEnvelope) {
...
 ) {
-  const query: TelemetryEventsQuery = { limit: params.limit ?? 100 };
+  // Pin the page to one absolute window so raw_replay cannot move as new
+  // telemetry arrives or relative time bounds age.
+  const now = new Date();
+  const query: TelemetryEventsQuery = {
+    limit: params.limit ?? 100,
+    until: params.until
+      ? resolveTelemetryTimeParam(params.until, now)
+      : now.toISOString(),
```
masnwilliams: equal query arguments do not make this an “exact page.” an offset-less `desc` replay can shift if a late event with `ts < until` arrives between calls, and retention can also change the archive. this also changes every existing no-`until` read to use the MCP server’s wall clock, which can exclude newest events if producer/API clocks differ. can the telemetry API return a snapshot/page cursor or both representations from one fetched page? otherwise preserve the old query semantics and describe replay as best-effort rather than exact.

src/lib/mcp/tools/browsers.ts
```diff
@@ -195,33 +199,74 @@ function compactTelemetryEvent({ seq, event }: TelemetryEnvelope) {
...
+  m: 60_000,
+  h: 3_600_000,
+} as const;
+
+const telemetryDurationPart =
+  /((?:\d+(?:\.\d*)?|\.\d+))(ns|us|µs|μs|ms|s|m|h)/g;
+const telemetryDuration =
+  /^[+-]?(?:0|(?:(?:\d+(?:\.\d*)?|\.\d+)(?:ns|us|µs|μs|ms|s|m|h))+)$/;
+
+function resolveTelemetryTimeParam(value: string, now: Date) {
```
masnwilliams: this is now a second implementation of the telemetry API’s duration grammar, including signs, compound units, and sub-millisecond values, but it converts through millisecond-resolution `Date`. that can drift from the server contract and silently alter valid bounds. i’d prefer the API to return normalized bounds or snapshot state. if that isn’t available, please move this into a canonical time utility with an injected clock and table-driven tests for all units, signs, compound values, invalid input, overflow, and fractional precision.

masnwilliams (changes requested):
the raw-read capability makes sense, but this needs a bounded output contract and a real replay consistency model before merging. the current implementation can emit very large MCP responses and calls equal query parameters an “exact page” even though the archive can mutate between reads. requesting changes on those two blockers.

the duration parser is also protocol duplication caused by the replay design; i’d prefer the API to own normalization and snapshot semantics. if local parsing remains, please isolate it and cover the grammar and precision edge cases.

src/lib/mcp/tools/browsers.ts
rgarcia: Addressed in 53ca6b7. Raw mode now requires explicit categories and limit <= 5 before calling the API, rejects screenshot categories and any returned PNG field, and rejects serialized responses over 1 MiB. Added regressions for missing/default limits, limits above 5, oversized multi-event pages, screenshot categories, and defensive PNG detection.

src/lib/mcp/tools/browsers.ts
rgarcia: Addressed in 53ca6b7. Removed the MCP-side wall-clock bound and restored the prior API query semantics for relative windows and offset-less desc reads. Replay is now named raw_replay_best_effort and documented as targeting the same page start while remaining subject to late events and retention.

src/lib/mcp/tools/browsers.ts
rgarcia: Addressed in 53ca6b7 by deleting the MCP duration parser entirely. Since/until values now pass through unchanged for the telemetry API to normalize, so there is no duplicate duration grammar or precision conversion in MCP.

masnwilliams (approved):
latest commit addresses the review findings: raw reads are bounded, replay is explicitly best-effort without changing API time semantics, and the duplicate duration parser is gone. the new boundary regressions pass, and CI and BugBot are green.

PR #159, opened 2026-08-15
Track client-declared MCP capabilities

rgarcia:
## summary

- add bounded capability properties to each `$mcp_initialize` analytics event
- distinguish sampling with tool support and elicitation form versus URL support
- track an explicit allowlist of official MCP extensions without retaining extension settings or unknown identifiers
- normalize legacy core task support with the Tasks extension

## validation

- `bun test`
- `bunx tsc --noEmit`
- `bunx prettier --check src/lib/mcp/analytics.ts src/lib/mcp/analytics.test.ts src/app/[transport]/route.ts`
- `bun run build` compiled and passed TypeScript, then stopped during page-data collection because `KERNEL_CLI_PROD_CLIENT_ID` is not available locally

---

> [!NOTE]
> **Low Risk**
> Analytics-only bounded booleans on initialize plus a stricter MCP Apps extension check; no auth or credential handling changes.
> 
> **Overview**
> **$mcp_initialize** events now include **bounded, allow-listed** properties derived from the client’s `initialize` capability map—sampling (and sampling tools), elicitation mode (`none` / `form` / `url` / `form_and_url`), legacy core `tasks`, and a fixed set of official MCP extensions (UI/apps, tasks, OAuth client credentials, enterprise auth). Unknown extension IDs and extension settings are **not** stored; malformed declarations are normalized to safe false/none values.
> 
> Capability parsing is **centralized** in `client-capabilities.ts` and reused by MCP Apps gating so an extension counts only when its entry is an object (e.g. `null` no longer counts as MCP Apps support). Integration tests assert these fields flow through instrumented analytics on initialize.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a9dd11421cbc068c49c35335a000e2bc143d5ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/app/[transport]/route.ts
+++ b/src/app/[transport]/route.ts
@@ -8,11 +8,9 @@ import { after, NextRequest } from "next/server";
 import { isValidJwtFormat } from "@/lib/auth-utils";
 import {
   captureMcpConnectionScopeFailure,
-  clientCapabilityAnalyticsFromInitialize,
   flushMcpAnalytics,
   instrumentMcpAnalytics,
   isMcpAnalyticsEnabled,
-  type McpClientCapabilityAnalytics,
 } from "@/lib/mcp/analytics";
 import {
   connectionAnalyticsFromContext,
@@ -132,7 +130,6 @@ async function handleMcpRequestWithIdentity({
   transportSessionId,
   connectionContextCacheIdentity,
   observeConnection,
-  clientCapabilityAnalytics,
 }: {
   req: NextRequest;
   token: string;
@@ -143,7 +140,6 @@ async function handleMcpRequestWithIdentity({
   transportSessionId: string | null;
   connectionContextCacheIdentity?: string;
   observeConnection: boolean;
-  clientCapabilityAnalytics: McpClientCapabilityAnalytics | null;
 }) {
   const [mcpApps, connection] = await Promise.all([
     requestUsesMcpApps(req, {
@@ -184,7 +180,6 @@ async function handleMcpRequestWithIdentity({
         ...authInfoExtra,
         connectionContext,
         connectionAnalytics,
-        clientCapabilityAnalytics,
       },
     }),
     {
@@ -199,7 +194,6 @@ async function handleAuthenticatedRequest(
   req: NextRequest,
   transportSessionId: string | null = null,
   observeConnection = false,
-  clientCapabilityAnalytics: McpClientCapabilityAnalytics | null = null,
 ): Promise<Response> {
   const authHeader = req.headers.get("Authorization");
   const token = authHeader?.startsWith("Bearer ")
@@ -225,7 +219,6 @@ async function handleAuthenticatedRequest(
       credentialType: "api_key",
       transportSessionId,
       observeConnection,
-      clientCapabilityAnalytics,
     });
   }
 
@@ -263,7 +256,6 @@ async function handleAuthenticatedRequest(
       ? `${authSubject}\0${transportSessionId}`
       : undefined,
     observeConnection,
-    clientCapabilityAnalytics,
   });
 }
 
@@ -294,9 +286,6 @@ export async function POST(req: NextRequest): Promise<Response> {
   }
   const isInitialize = parsed?.method === "initialize";
   const initializeParams = isInitialize ? parsed?.params : undefined;
-  const clientCapabilityAnalytics = isInitialize
-    ? clientCapabilityAnalyticsFromInitialize(parsed)
-    : null;
   const isStreamableInitialize =
     new URL(req.url).pathname.endsWith("/mcp") && isInitialize;
   const session = isStreamableInitialize
@@ -321,7 +310,6 @@ export async function POST(req: NextRequest): Promise<Response> {
     }),
     session?.id ?? null,
     isInitialize,
-    clientCapabilityAnalytics,
   );
 
   if (!session) return response;
--- a/src/app/[transport]/route.ts
+++ b/src/app/[transport]/route.ts
@@ -8,9 +8,11 @@ import { after, NextRequest } from "next/server";
 import { isValidJwtFormat } from "@/lib/auth-utils";
 import {
   captureMcpConnectionScopeFailure,
+  clientCapabilityAnalyticsFromInitialize,
   flushMcpAnalytics,
   instrumentMcpAnalytics,
   isMcpAnalyticsEnabled,
+  type McpClientCapabilityAnalytics,
 } from "@/lib/mcp/analytics";
 import {
   connectionAnalyticsFromContext,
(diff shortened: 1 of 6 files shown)


src/lib/mcp/analytics.ts
```diff
@@ -141,6 +181,63 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+
+  const properties: McpClientCapabilityAnalytics = {
+    [MCP_CLIENT_SUPPORTS_SAMPLING_PROPERTY]: sampling !== null,
+    [MCP_CLIENT_SUPPORTS_SAMPLING_TOOLS_PROPERTY]:
+      sampling !== null && hasOwn(sampling, "tools"),
+    [MCP_CLIENT_ELICITATION_MODE_PROPERTY]: elicitation
+      ? hasOwn(elicitation, "url")
+        ? "form_and_url"
+        : "form"
+      : "none",
```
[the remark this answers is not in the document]
rgarcia: fixed in c30eec0: URL-only elicitation now records `url`, while empty elicitation remains implicit `form` and explicit form + URL records `form_and_url`. added coverage for all three cases.

src/app/[transport]/route.ts
```diff
@@ -286,6 +294,9 @@ export async function POST(req: NextRequest): Promise<Response> {
   }
   const isInitialize = parsed?.method === "initialize";
   const initializeParams = isInitialize ? parsed?.params : undefined;
+  const clientCapabilityAnalytics = isInitialize
+    ? clientCapabilityAnalyticsFromInitialize(parsed)
```
Comment (masnwilliams) [inline src/app/[transport]/route.ts]: this pushes an analytics-only DTO through the route and both authentication helpers even though `eventProperties` already receives the raw initialize request. that adds parameters to unrelated auth boundaries, stores the DTO in `authInfo.extra`, and casts it back out in analytics just to recover data already available as `request`. please call `clientCapabilityAnalyticsFromInitialize(request)` directly inside `eventProperties` and delete this route/auth plumbing plus `clientCapabilityAnalyticsContext`. i tested that restructuring locally: it removes 24 lines and passes the analytics/route tests, `tsc`, and Prettier. the integration test can then stop precomputing and injecting the same value, so it exercises the real request-to-event boundary instead of mirroring the implementation.

src/lib/mcp/analytics.ts
```diff
@@ -141,6 +183,70 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+    [MCP_CLIENT_SUPPORTS_APPS_PROPERTY]: false,
+    [MCP_CLIENT_SUPPORTS_TASKS_PROPERTY]: hasOwn(capabilities, "tasks"),
+    [MCP_CLIENT_SUPPORTS_OAUTH_CLIENT_CREDENTIALS_PROPERTY]: false,
+    [MCP_CLIENT_SUPPORTS_ENTERPRISE_AUTH_PROPERTY]: false,
+  };
+
+  for (const [extension, property] of Object.entries(
+    CLIENT_EXTENSION_PROPERTIES,
+  )) {
+    if (extensions && hasOwn(extensions, extension))
```
masnwilliams: `hasOwn` treats malformed settings such as `null`, `false`, strings, or arrays as extension support. the extension contract requires a settings object; `{}` is the valid presence-only declaration. this matters with the pinned SDK because unknown `extensions` data is stripped rather than rejected, so an initialize payload with `"io.modelcontextprotocol/ui": null` is accepted while this reducer records Apps support as `true`. please require an object-valued declaration (for example `isRecord(extensions[extension])`) and add malformed-value coverage.

src/lib/mcp/analytics.ts
```diff
@@ -68,6 +68,41 @@ const posthog = projectToken
...
+  [MCP_CLIENT_SUPPORTS_APPS_PROPERTY]: boolean;
+  [MCP_CLIENT_SUPPORTS_TASKS_PROPERTY]: boolean;
+  [MCP_CLIENT_SUPPORTS_OAUTH_CLIENT_CREDENTIALS_PROPERTY]: boolean;
+  [MCP_CLIENT_SUPPORTS_ENTERPRISE_AUTH_PROPERTY]: boolean;
+};
+
+// Official extensions listed at https://modelcontextprotocol.io/extensions.
+// Keep this explicit: arbitrary extension identifiers and settings must not enter analytics.
+const CLIENT_EXTENSION_PROPERTIES = {
+  "io.modelcontextprotocol/ui": MCP_CLIENT_SUPPORTS_APPS_PROPERTY,
```
masnwilliams: this creates another source of truth for the Apps extension identifier and initialize-capability parsing alongside `MCP_APPS_EXTENSION` / `initializeDeclaresMcpApps` in `mcp-apps-gate.ts`. the two paths already disagree on malformed values, and the locally reimplemented elicitation semantics required a corrective follow-up commit. please centralize the raw client-capability model and official extension identifiers in a focused module used by both the gate and analytics (and reuse the SDK's elicitation helper where practical) so these protocol rules cannot drift independently.

masnwilliams (changes requested):
requesting changes for one structural blocker and one capability-boundary correctness issue. the analytics reducer itself is appropriately bounded, and the full test suite, TypeScript, and formatting checks pass, but the implementation should keep analytics in the instrumentation layer and validate extension declarations before these properties become durable telemetry.

src/app/[transport]/route.ts
Comment (rgarcia) [reply src/app/[transport]/route.ts]: fixed in 7a9dd11: capability analytics now derives directly from the raw initialize request inside `eventProperties`. removed the route/auth parameters, `authInfo.extra` DTO, and analytics context cast; the integration test now exercises the request-to-event path.

Comment (rgarcia) [reply src/app/[transport]/route.ts]: fixed in 7a9dd11: capability analytics now derives directly from the raw initialize request inside `eventProperties`. removed the route/auth parameters, `authInfo.extra` DTO, and analytics context cast; the integration test now exercises the request-to-event path.

src/lib/mcp/analytics.ts
rgarcia: fixed in 7a9dd11: extension support now requires an object-valued declaration. added coverage for null, boolean, string, and array settings, including the Apps gate.

src/lib/mcp/analytics.ts
rgarcia: fixed in 7a9dd11: added `client-capabilities.ts` as the shared raw capability boundary and moved all official extension identifiers there. both Apps gating and analytics use its parsing/validation, and elicitation mode normalization delegates to the SDK helper.

masnwilliams (approved):
latest commit addresses all blocking feedback: capability analytics now stays in eventProperties, malformed declarations fail closed, and capability parsing plus extension identifiers are centralized and shared with the MCP Apps gate. verified locally: 242 tests pass, TypeScript passes, and changed files pass Prettier. no remaining findings.

rgarcia:
cursor review

rgarcia:
bugbot run verbose=true

PR #161, opened 2026-08-20
Add feedback reporting tool

rgarcia:
## summary

- add a `submit_feedback` tool for product, mcp, documentation, and general feedback
- record submissions as `mcp_feedback_submitted` PostHog events with organization attribution
- redact obvious emails, URLs, and token-shaped values before capture
- keep feedback failures non-fatal and document the new standalone tool

## testing

- `bun test`
- `bunx tsc --noEmit`
- `bunx prettier --check README.md src/lib/mcp/analytics.ts src/lib/mcp/analytics.test.ts src/lib/mcp/tools/feedback.ts src/lib/mcp/tools/feedback.test.ts`
- `KERNEL_CLI_PROD_CLIENT_ID=test-prod KERNEL_CLI_STAGING_CLIENT_ID=test-staging KERNEL_CLI_DEV_CLIENT_ID=test-dev bun run build`

### DIFF
--- a/src/lib/mcp/analytics.ts
+++ b/src/lib/mcp/analytics.ts
@@ -13,6 +13,10 @@ import type {
   McpConnectionContext,
 } from "@/lib/mcp/auth-context";
 import {
+  type KernelFeedback,
+  registerFeedbackTool,
+} from "@/lib/mcp/tools/feedback";
+import {
   clientDeclaresExtension,
   clientElicitationModes,
   initializeClientCapabilities,
@@ -60,6 +64,7 @@ export type McpConnectionScopeFailureAnalytics = {
 
 export const MCP_CONNECTION_SCOPE_FAILURE_EVENT =
   "mcp_connection_scope_failure";
+export const MCP_FEEDBACK_SUBMITTED_EVENT = "mcp_feedback_submitted";
 
 if (!projectToken && process.env.NODE_ENV !== "production") {
   console.error(
@@ -262,13 +267,16 @@ function annotateProjectParamUsage(properties: Record<string, unknown>) {
   properties[MCP_USED_PROJECT_PROPERTY] = hasNonEmptyParam(args, "project");
 }
 
-function sanitizeIntent(intent: string) {
-  const redacted = INTENT_REDACTIONS.reduce(
-    (text, [pattern, replacement]) => text.replace(pattern, replacement),
-    intent.trim(),
+function redactAnalyticsText(text: string) {
+  return INTENT_REDACTIONS.reduce(
+    (redacted, [pattern, replacement]) =>
+      redacted.replace(pattern, replacement),
+    text.trim(),
   );
+}
 
-  return redacted.slice(0, INTENT_MAX_LENGTH);
+function sanitizeIntent(intent: string) {
+  return redactAnalyticsText(intent).slice(0, INTENT_MAX_LENGTH);
 }
 
 // Must stay the SDK's default name: reportMissing advertises a tool under this name and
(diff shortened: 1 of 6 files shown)


src/lib/mcp/analytics.ts
```diff
@@ -486,11 +499,61 @@ export function captureMcpConnectionScopeFailure(
...
+  extra: unknown,
+  client: PostHog | null = posthog,
+) {
+  if (!client) return;
+
+  const organizationId = connectionOrgId(extra);
+  const userId = connectionUserId(extra);
+
+  client.capture({
+    distinctId: userId ?? "mcp-feedback",
```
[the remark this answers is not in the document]
rgarcia: fixed in a8705c1. feedback events now use the verified MCP transport session ID, fall back to the SDK-derived transport session ID, and mint a fresh anonymous session ID only when neither is available. added coverage for transport attribution and fallback uniqueness.

src/lib/mcp/tools/feedback.ts
```diff
@@ -0,0 +1,153 @@
...
+  capture: KernelFeedbackCapture,
+) {
+  server.registerTool(
+    KERNEL_FEEDBACK_TOOL_NAME,
+    {
+      title: "submit KERNEL feedback",
+      description: TOOL_DESCRIPTION,
+      inputSchema: feedbackFields,
+      annotations: {
+        readOnlyHint: true,
```
masnwilliams: `submit_feedback` creates an external analytics event, so advertising it as read-only is inaccurate for MCP clients that use these hints for approval or policy decisions. Please set `readOnlyHint: false`; the existing `destructiveHint: false` and `idempotentHint: false` values look right.

src/lib/mcp/tools/feedback.ts
```diff
@@ -0,0 +1,153 @@
...
+        destructiveHint: false,
+        idempotentHint: false,
+        openWorldHint: false,
+      },
+    },
+    async ({ context: _context, ...feedback }, extra) => {
+      try {
+        await capture(feedback, extra);
+      } catch {
```
masnwilliams: This path always returns `received: true` and says the feedback “has been recorded,” even when capture throws. The same false confirmation occurs when analytics is unavailable because `instrumentMcpAnalytics` installs a no-op capture callback. Keeping feedback failures non-blocking makes sense, but the response should only claim recording after a successful capture and should otherwise return an explicit unavailable/failure status.

masnwilliams (changes requested):
two small correctness issues to address. the overall implementation and test coverage look good; I’m not treating the free-form analytics payload or global registration as blockers.

src/lib/mcp/tools/feedback.ts
rgarcia: fixed in f108770. successful capture returns `recorded: true` with `status: "recorded"`; capture exceptions return `recorded: false` / `status: "failed"`; disabled analytics returns `recorded: false` / `status: "unavailable"`. all paths still tell the agent to continue the original task, and tests cover each outcome.

src/lib/mcp/tools/feedback.ts
rgarcia: fixed in f108770. `submit_feedback` now advertises `readOnlyHint: false`; the destructive and idempotent hints remain false.

masnwilliams (approved):
latest commit addresses both findings: feedback now reports recorded, failed, and unavailable outcomes accurately, and the tool is correctly marked non-read-only. focused tests and type-check pass; CI and BugBot are green.

PR #165, opened 2026-08-24
Record the OAuth Redis persistence contract as golden fixtures

rgarcia:
## Summary

The OAuth /token and /authorize endpoints persist org/project authorization context to Redis, and other services read those exact keys and values back on every authenticated request. Today nothing pins that contract: a change to a key name, HMAC derivation, serialization shape, or TTL can merge silently and remap existing tokens.

This PR adds a recording harness that drives the real request handlers (`tokenRequest`, `authorizeRequest`) through their existing dependency-injection seams with pinned fake Clerk responses (exchange, verify, membership) and fake auth — persistence itself is untouched production code writing to a real Redis. Every key written is dumped with its value and TTL to a committed JSON fixture whose inputs are all fixed (secret, client ids, PKCE verifier, token strings), so re-runs are byte-identical.

CI regenerates the fixture and fails when it differs from what's committed, so any change to the Redis contract now shows up as a reviewable diff.

- `scripts/record-oauth-redis-contract.ts` — harness; runs against an isolated disposable Redis (default db 15, flushed between scenarios)
- `fixtures/oauth-redis-recordings.json` — the recordings plus self-documentation: pinned env values, identifiers, PKCE pair, per-flow Clerk response bodies, and key-derivation notes sufficient for a foreign implementation to replay
- `.github/workflows/ci.yml` — redis service container + regeneration/diff enforcement step

## Scenarios recorded

1. `authorization_code_org_scoped_pkce` — authorize writes `oauth-request:<hmac(client_id:code_challenge)>`; the exchange consumes it and persists `jwt:`/`refresh:` mappings
2. `authorization_code_project_scoped_pkce` — same, with `project_id` in the persisted context
3. `refresh_token_rotation_sliding_ttl` — old `refresh:` key deleted, rotated key written with the full sliding TTL
4. `legacy_non_pkce_client` — allowlisted non-PKCE client stores under literal `client:<client_id>`
5. `refresh_backfills_clerk_user_id_from_id_token` — legacy context without `clerk_user_id` gets it backfilled from the id_token subject on refresh

## Test plan

- [x] `bun test`: 248 pass
- [x] `bunx tsc --noEmit` clean; prettier clean on changed files
- [x] Harness run twice consecutively produces byte-identical fixtures
- [x] Recorded HMAC keys independently verified against the documented derivations

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,16 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    services:
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 20
     steps:
       - uses: actions/checkout@v4
 
@@ -28,3 +38,11 @@ jobs:
 
       - name: Test
         run: bun test
+
+      - name: OAuth Redis recordings are up to date
+        run: |
+          bun scripts/record-oauth-redis-contract.ts
+          git diff --exit-code -- fixtures/oauth-redis-recordings.json \
+            || (echo "fixtures/oauth-redis-recordings.json differs from what the OAuth code now writes." \
+                echo "Re-run: bun scripts/record-oauth-redis-contract.ts, review the diff, and commit it." \
+                exit 1)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,8 @@ jobs:
       - name: OAuth Redis recordings are up to date
         run: |
           bun scripts/record-oauth-redis-contract.ts
-          git diff --exit-code -- fixtures/oauth-redis-recordings.json \
-            || (echo "fixtures/oauth-redis-recordings.json differs from what the OAuth code now writes." \
-                echo "Re-run: bun scripts/record-oauth-redis-contract.ts, review the diff, and commit it." \
-                exit 1)
+          git diff --exit-code -- fixtures/oauth-redis-recordings.json || {
+            echo "fixtures/oauth-redis-recordings.json differs from what the OAuth code now writes."
+            echo "Re-run: bun scripts/record-oauth-redis-contract.ts, review the diff, and commit it."
+            exit 1
+          }
--- a/scripts/record-oauth-redis-contract.ts
+++ b/scripts/record-oauth-redis-contract.ts
@@ -111,6 +111,23 @@ interface Snapshot {
   keys: { key: string; value: string; ttl_seconds: number }[];
 }
 
+// Every TTL this code writes is one of the pinned constants. Reading the
+// TTL back can land a second low depending on timing, so snap values within
+// tolerance to their constant and treat anything else as a contract change.
+const PINNED_TTL_SECONDS = [EXPIRES_IN_SECONDS, REFRESH_TOKEN_ORG_TTL_SECONDS];
+
+function pinnedTtl(seconds: number): number {
+  const pinned = PINNED_TTL_SECONDS.find(
+    (value) => Math.abs(value - seconds) <= 5,
+  );
+  if (!pinned) {
+    throw new Error(
+      `Redis TTL ${seconds}s matches no pinned TTL (${PINNED_TTL_SECONDS.join("s, ")}s)`,
+    );
+  }
+  return pinned;
+}
+
 async function snapshot(after: string): Promise<Snapshot> {
   const keys = (await redisClient.keys("*")).sort();
   return {
@@ -119,7 +136,7 @@ async function snapshot(after: string): Promise<Snapshot> {
       keys.map(async (key) => ({
         key,
         value: (await redisClient.get(key))!,
-        ttl_seconds: await redisClient.ttl(key),
+        ttl_seconds: pinnedTtl(await redisClient.ttl(key)),
       })),
     ),
   };
@@ -179,7 +196,7 @@ async function authorize(query: Record<string, string>): Promise<void> {
   const response = await authorizeRequest(request, authorizeDependencies());
   if (response.status !== 307) {
     throw new Error(
-      `authorize redirected to ${response.status}: ${await response.text()}`,
+      `authorize returned ${response.status}, expected 307 redirect: ${await response.text()}`,
     );
   }
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -28,3 +38,11 @@ jobs:
...
       - name: Test
         run: bun test
+
+      - name: OAuth Redis recordings are up to date
+        run: |
+          bun scripts/record-oauth-redis-contract.ts
+          git diff --exit-code -- fixtures/oauth-redis-recordings.json \
+            || (echo "fixtures/oauth-redis-recordings.json differs from what the OAuth code now writes." \
+                echo "Re-run: bun scripts/record-oauth-redis-contract.ts, review the diff, and commit it." \
+                exit 1)
```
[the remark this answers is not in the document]
rgarcia: Fixed in 92b80f8 — the failure branch now uses an explicit `{ ... }` block with newline separators so `exit 1` actually executes. Verified both directions with the step's exact commands: clean tree → PASS (exit 0), corrupted committed fixture → FAIL (exit 1) with both messages printed, then restored → PASS. Real CI on this commit also passes the regenerated-diff gate against a live redis service.

scripts/record-oauth-redis-contract.ts
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// Every TTL this code writes is one of the pinned constants. Reading the
+// TTL back can land a second low depending on timing, so snap values within
+// tolerance to their constant and treat anything else as a contract change.
+const PINNED_TTL_SECONDS = [EXPIRES_IN_SECONDS, REFRESH_TOKEN_ORG_TTL_SECONDS];
+
+function pinnedTtl(seconds: number): number {
+  const pinned = PINNED_TTL_SECONDS.find(
+    (value) => Math.abs(value - seconds) <= 5,
```
masnwilliams: **blocker:** This tolerance hides real TTL contract changes. I changed both authorization-request TTL writes from 3600s to 3598s and the regenerated fixture remained byte-identical, so the CI gate would pass even though production behavior changed. Please capture the exact TTL passed through the existing dependency seams and record that value; use Redis TTL only to verify that expiration was applied.

scripts/record-oauth-redis-contract.ts
```diff
@@ -0,0 +1,437 @@
...
+// Requires a disposable Redis: every key in the selected database is
+// flushed between scenarios. Point it at an isolated instance with
+// OAUTH_RECORDING_REDIS_URL (default redis://127.0.0.1:6379/15).
+
+// Hard-assign every environment input the OAuth code reads instead of
+// inheriting ambient values: recordings must be reproducible, and an
+// inherited REDIS_URL could point the flushes below at real data.
+const recordingRedisUrl =
+  process.env.OAUTH_RECORDING_REDIS_URL ?? "redis://127.0.0.1:6379/15";
+if (!new URL(recordingRedisUrl).pathname.endsWith("/15")) {
```
masnwilliams: **blocker:** Selecting database 15 does not establish isolation: any shared or production Redis URL ending in `/15` passes this check, and the script then calls `FLUSHDB` repeatedly. Please reject non-loopback hosts (or otherwise require a verifiably disposable instance), and put the final flush/disconnect in `finally` so failed recordings do not leave fixture keys behind.

masnwilliams (changes requested):
The recording approach is useful, but these two issues undermine its exact drift-detection guarantee and make the recorder unsafe outside the default local setup. Requesting changes before merge.

scripts/record-oauth-redis-contract.ts
rgarcia: Fixed in 01cb5cb. The harness now wraps the injected writers (persistOAuthTokenContexts, setAuthorizationContextForRequest, setAuthorizationContextForClientId) and records the exact ttlSeconds arguments production passes, matched to each discovered key by stored value; the Redis TTL read is only a sanity check that expiration was applied (10s tolerance). Re-ran your experiment: changing the authorization-request TTL writes from 60*60 to 3598 now regenerates a fixture with `ttl_seconds: 3598` on those keys — 12 lines diff, CI would fail. With production constants restored, the recorded contract values are unchanged from before; only the fixture's meta descriptions changed.

scripts/record-oauth-redis-contract.ts
rgarcia: Fixed in 01cb5cb. Non-loopback hosts are now rejected unless OAUTH_RECORDING_ALLOW_REMOTE_REDIS=1 is exported (the db-15 check stays as defense in depth), and the final flush + disconnect moved into a nested finally so a failed recording still cleans up. Verified: a redis://shared-host:6379/15 URL is refused with an error naming the escape hatch.

masnwilliams (changes requested):
The Redis isolation and cleanup blocker is addressed. The TTL blocker remains: snapshots match the first captured write with the same serialized value and a live TTL within 10 seconds, so an authorization-code JWT write can be attributed to the earlier /authorize capture. I reproduced this by changing only authorization-code JWT persistence from `expires_in` to `expires_in - 2`; the regenerated fixture remained byte-identical. Please associate captured writes with their exact Redis keys rather than matching by value.

scripts/record-oauth-redis-contract.ts
```diff
@@ -0,0 +1,499 @@
...
+  keys: { key: string; value: string; ttl_seconds: number }[];
+}
+
+// TTLs are captured where production code hands them to the real writers,
+// keyed by the exact Redis key each writer derives, so a recorded TTL is
+// definitionally the TTL requested for that key — writes carrying the same
+// serialized value (e.g. oauth-request vs jwt vs refresh) can never be
+// attributed to each other. The live Redis TTL is read back only as a
+// sanity check that expiration was applied.
+const capturedTtlByKey = new Map<string, number>();
```
rgarcia, line 133: Replying to masnwilliams's second-round blocker: attribution now goes through exact Redis keys. The seam wrappers recompute the writers' derivations (jwt:/refresh:/oauth-request: HMACs over the pinned CLERK_SECRET_KEY, literal client:<id>) and record ttlSeconds per key in capturedTtlByKey — writes carrying identical serialized values can no longer be attributed to each other. Verified your repro: jwt persistence changed to expires_in - 2 diffs exactly the 7 jwt keys to 3598 while refresh/oauth-request/client stay 3600; the earlier authorize 3600→3598 repro also still diffs its own keys only. The 10s tolerance remains solely as an expiration sanity check per key.

masnwilliams (approved):
Both blockers are addressed. TTL captures are now attributed to exact Redis keys, and the authorization-code-only `expires_in - 2` mutation produces a fixture diff. Redis isolation and failure cleanup remain in place. Tests, type-checking, formatting, CI, and BugBot are green.

PR #167, opened 2026-08-26
Fix managed auth telemetry JSON Schema refs

masnwilliams:
## summary

- generate independent schemas for each managed-auth browser telemetry category
- prevent MCP `tools/list` from emitting local `$ref` entries that some JSON Schema consumers cannot resolve
- add a regression test against the published tool schema

## testing

- `bun test`
- `bunx tsc --noEmit`
- `bunx prettier --check src/lib/mcp/tools/auth-connections.test.ts src/lib/mcp/tools/managed-auth-telemetry.ts`
- reproduced with Zod 4.4.3 `fromJSONSchema` before the fix and verified conversion succeeds after it

---

> [!NOTE]
> **Low Risk**
> Schema-shape refactor with unchanged validation rules (`superRefine` preserved) plus a regression test; no runtime auth or API behavior changes.
> 
> **Overview**
> Fixes MCP `tools/list` output for **`manage_auth_connections`** (and related tools using the same schema) so **`browser_telemetry`** no longer embeds unresolved local **`$ref`** entries.
> 
> **`telemetryCategorySchema`** is now a factory that returns a fresh Zod object for each browser category (captcha, network, page, etc.) instead of reusing one shared schema. Zod’s JSON Schema conversion was deduplicating that shared shape into **`$ref`**s, which broke consumers such as Zod 4’s **`fromJSONSchema`**.
> 
> Adds an integration test that lists MCP tools and asserts the published **`browser_telemetry`** property JSON does not contain **`"$ref"`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f77cb64eb1426ef3add433fc318938838638e009. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/mcp/tools/auth-connections.test.ts
+++ b/src/lib/mcp/tools/auth-connections.test.ts
@@ -1,5 +1,7 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { organizationWideAuthInfo } from "@/lib/mcp/auth-context.test-fixtures";
+import { connectTestMcp } from "@/lib/mcp/mcp-test-fixtures";
+import { registerAuthConnectionTools } from "./auth-connections";
 import { toSafeAuthConnection } from "./managed-auth-state";
 import {
   assertNoSecrets,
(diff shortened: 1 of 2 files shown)


dcruzeneil2 (approved):
looks good. the change removes the problematic `$ref`s without changing validation or runtime behavior, and the regression test covers the published MCP schema. full test suite and type-check pass. approved.

PR #186, opened 2026-09-10
Support vault provider configs and imported Link wallets

rgarcia:
## SDK dependency

Uses the published **`@onkernel/sdk` 0.101.0** (`^0.101.0` in package.json, pinned in bun.lock), released from commit `88e5ccf02ff0b653f8951d38a508a684fe12231b`. The stable-SDK merge gate is resolved: the branch now installs, typechecks, and builds using its committed lockfile without a staging dependency or local override.

The released vault resource sources match the previously validated preview `ae29b778cecc8aadbf1922c29cd1fcb0464f9ea6`. No generated SDK source is committed. Deployment still requires the new API routes and the appropriate vault entitlement; deployed availability was not tested.

## Summary

- Add consolidated `manage_vault_provider_configs` create/list/get/update/delete actions for organization-owned Link and AgentCard configs, with organization-scoped write checks, single-page pagination, secret rotation without identity changes, no retries, and existing deletion/error conventions.
- Extend wallet creation with config ID/name selection and backend-supplied Link grants while retaining Kernel-managed defaults. Preserve resolved bindings in public output; distinguish application credentials from user grants and document immutable grants and refresh ownership.
- Keep secrets out of public projections, validation failures, upstream errors, SDK payload logs, and analytics. Sanitize validation failures at registration while advertising strict schemas. Sanitize complete responses and omit executable hints containing supplied secrets, including collisions in target identifiers.
- Update card lifecycle guidance for pending edits, omitted versus empty fields, immutable bindings, and `recovery_required`. Existing user-approval, entitlement gating, project scoping, and no-payment-retry behavior remain intact.

## Validation

Rerun against the published 0.101.0 package after `bun install --force --frozen-lockfile`:

- `bun test`: **517 passed, 0 failed** across 51 files, using mocked transports; covers tool routing, schemas, scopes, pagination, redaction, Kernel-managed compatibility, and recovery behavior.
- `bunx tsc --noEmit`: passed.
- `bun run build`: passed with dummy build-time OAuth/Clerk configuration; no real authentication or provider operation.
- Changed-file Prettier and `git diff --check`: passed.

Without build-time OAuth configuration, Next.js page-data collection fails on existing required client IDs. The existing `next lint` script is unsupported on Next.js 16; changed-file Prettier was used instead.

No live provider calls, production mutations, deployment, merge, or release performed by this PR. Review fixes add generic shared 403 guidance and preserve flat multi-action tool handling of unused fields, with regression coverage. Human approval and latest-commit checks are required before merge.


src/lib/mcp/vault-responses.ts
```diff
@@ -192,7 +225,14 @@ const vaultErrorMessages = new Map([
...
   ],
-  ["not_found", "Vault, item, or project not found or unavailable."],
+  [
+    "not_found",
+    "Vault, item, provider configuration, or project not found or unavailable.",
+  ],
+  [
+    "forbidden",
+    "This credential cannot perform the vault operation. Provider configuration writes require organization scope.",
+  ],
```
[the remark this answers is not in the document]
rgarcia: Fixed in cbe23a3: the shared 403 message now asks callers to check connection scope and permissions, without implying a provider-config write. Added a project-owned item regression test. The config tool retains its specific local organization-scope check.

src/lib/mcp/tools/vault-provider-configs.ts
```diff
@@ -0,0 +1,206 @@
...
+        );
+      }
+      if (
+        !["create", "update"].includes(params.action) &&
+        (params.credentials !== undefined || params.name !== undefined)
+      ) {
+        return errorResponse(
+          "name and credentials are only accepted for create or update.",
+        );
+      }
```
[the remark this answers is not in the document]
rgarcia: Fixed in cbe23a3: removed the provider/name/credentials action-to-field guards. The switch ignores unused fields and sends only the selected action’s body. Added real MCP transport tests for list/get/delete with unused write fields and update with a create-only provider. Organization-scoped writes and immutable client_id validation are preserved.

src/lib/mcp/vault-responses.ts
```diff
@@ -192,7 +225,14 @@ const vaultErrorMessages = new Map([
...
     "Invalid vault request. Check the tool's documented inputs.",
   ],
-  ["not_found", "Vault, item, or project not found or unavailable."],
+  [
+    "not_found",
+    "Vault, item, provider configuration, or project not found or unavailable.",
+  ],
+  [
+    "forbidden",
+    "This credential cannot perform the vault operation. Provider configuration writes require organization scope.",
```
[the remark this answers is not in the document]
rgarcia: Applied in cbe23a3 and covered with a 403 regression test for project-owned vault items.

src/lib/mcp/vault-schemas.ts
```diff
@@ -41,29 +41,91 @@ export const vaultWaitSchema = z
...
+            })
+            .strict("Unexpected OAuth client fields."),
+          tokens: secretInputObject({
+            access_token: z.string().min(1),
+            refresh_token: z.string().min(1),
+          }).describe(
+            "Write-only token pair from the same grant. Supply through a trusted backend, never chat. Kernel owns subsequent refresh rotation.",
+          ),
+        })
+        .strict("Unexpected imported authorization fields."),
```
hiroTamada: suggestion (non-blocking): sanitize validation failures at the vault tool boundary rather than maintaining separate safe/unsafe object schemas. `.strict("…")` changes the message but still includes raw unknown keys in the Zod error. an otherwise valid imported wallet with an extra `authorization["review-secret-sentinel"]` property returns that sentinel in the MCP error's `keys` array, before `throwVaultError` runs.

using ordinary strict schemas behind a sanitized validation boundary would also remove the `secretInputObject` workaround, whose `.passthrough().refine(...)` currently advertises `additionalProperties: true` in tools/list while rejecting extras at runtime. worth covering misplaced secrets at each nesting level with transport tests.

src/lib/mcp/vault-responses.ts
```diff
@@ -163,8 +188,15 @@ export function vaultObservationHints(target: VaultItemTarget, after?: string) {
...
 }
 
-export function vaultItemResponse(item: unknown, target: VaultItemTarget) {
-  const projected = projectVaultOutput(item, vaultItemFields);
+export function vaultItemResponse(
+  item: unknown,
+  target: VaultItemTarget,
+  secrets: (string | undefined)[] = [],
+) {
+  const projected = redactVaultSecrets(
```
hiroTamada: suggestion (non-blocking): apply supplied-secret protection to the complete response, not just the projected item. the hints here are constructed afterward from unsanitized target fields. in a mocked wallet import where a synthetic access token was also accidentally supplied as the wallet key, `item.key` became `[redacted]` but both observation hints returned the raw token in `arguments.key`.

keep the upstream field allowlist, then sanitize the complete response at one output boundary. omit unsafe executable hints rather than replacing their identifiers with unusable redacted values. this would also simplify the current project → stringify → parse → construct more output → stringify sequence.

hiroTamada (approved):
approved with two non-blocking suggestions for consolidating secret handling at the validation and response boundaries.

validation: 500 tests passed and `bunx tsc --noEmit` passed. the edge cases in the inline suggestions were reproduced using the real MCP in-memory transport with mocked upstream responses.

src/lib/mcp/vault-schemas.ts
rgarcia: Addressed in 6e78833. All five vault tools now wrap each input field at registration with a preprocessing boundary that emits only a fixed validation issue on failure. The ordinary strict schemas still drive tools/list and successful callback parsing; secretInputObject and its passthrough workaround are removed. Real MCP transport regressions cover rejected keys at spec/authorization/client/provider_config/tokens depth, rejected discriminator values, config validation, and additionalProperties: false in discovery. Full suite: 517 passed; typecheck and production build passed. Please re-review this commit before merge.

src/lib/mcp/vault-responses.ts
rgarcia: Addressed in 6e78833. Public projection is followed by complete-response sanitization with one final serialization, without a stringify/parse roundtrip. Observation and invocation hints containing supplied secrets are omitted, never given redacted executable identifiers. The config tool also sanitizes complete acknowledgments, including delete/404 responses. Transport tests reproduce key/vault/project collisions, unsafe operation names, retained safe hints, and config selector collisions.

hiroTamada (approved):
reviewed incrementally — approved. validation: 517 tests passed and TypeScript checking passed. no live provider testing.

PR #197, opened 2026-09-16
Preserve legacy capability report calls

masnwilliams:
## summary
- accept context-only `get_more_tools` calls from cached clients as a non-recording refresh response
- keep incomplete structured reports out of capability-demand analytics without failing the tool call
- classify genuine input-schema rejections as validation errors

## tests
- `bun test`
- `bun x tsc --noEmit`
- changed-file Prettier check
- `bun run build`

---

> [!NOTE]
> **Low Risk**
> Behavior is narrowly scoped to exact legacy payloads and analytics labeling; structured reporting and demand capture paths are unchanged.
> 
> **Overview**
> Adds **backward compatibility** for clients still calling `get_more_tools` with only a `context` argument after the structured schema rollout. Those calls are intercepted at the `tools/call` layer (before strict schema validation) and return a successful **`legacy_schema_refresh_required`** payload with **`recorded: false`**, nudging a tools refresh instead of failing the agent. Partially structured payloads still validate against the full schema and error as before.
> 
> **`get_more_tools`** is registered via **`registerTool`** with the structured Zod input unchanged. Analytics registration order is flipped so instrumentation wraps the legacy shim and still emits tool-call telemetry for compatibility responses. PostHog **`$mcp_tool_call`** events for genuine schema rejections on this tool are reclassified from generic **`Error`** to **`validation`**.
> 
> Docs and tests cover the legacy path, partial-input rejection, and validation telemetry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdca031b0a034488cf498792030fbd27a47071a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rgarcia (approved):
reviewed — the compatibility paths are clear and the targeted tests pass. two things worth addressing:

### Questions

- `src/lib/mcp/tools/missing-capability.ts:112-119` — the refreshed schema also advertises every structured field as optional. a compliant current client can repeatedly send only `context`, receive `legacy_schema_refresh_required`, refresh to the same schema, and never record demand. could we preserve the required fields in the advertised current contract while accepting cached legacy calls at dispatch, or otherwise distinguish refreshed clients?

### Test coverage

- `src/lib/mcp/analytics.test.ts:346-371` — this test constructs the exact SDK error shape expected by the classifier, so it cannot catch changes in the real instrumentation path. consider sending an invalid `get_more_tools` call through the existing `simulateRequest` harness and asserting the captured `$mcp_error_type` is `validation`.

masnwilliams:
addressed both review points in 115f0fd:

- restored all structured fields as required in the advertised schema; the exact legacy context-only payload is handled at dispatch without changing `tools/list`
- replaced the synthetic classifier test with an invalid call through the instrumented request harness and asserted the captured validation type plus error-text redaction

validated with 606 tests, typecheck, formatting, and a production build.

PR #201, opened 2026-09-17
Make the execution ladder legible from every rung

robertjamesprior:
Second half of the metadata work. #198 covered the catalog entry and the server `instructions`; this covers the tool descriptions, where the model actually chooses.

## The gap

`webmcp` already handles this well. Its description tells the caller what to do when a site offers no tools of its own: *"use execute_playwright_code or computer_action instead of invoking."*

The other three rungs say nothing about each other. `manage_browsers` creates the session every web task runs in and never mentions how to drive it. `execute_playwright_code` points at `manage_browsers` for lifecycle but not at the layers either side of it. `computer_action` opened with "Execute computer actions on a browser session", which tells a model nothing it did not already know from the tool name.

So a model choosing between them has to already know the order, which is exactly what we cannot assume.

## The change

Three opening sentences. Operational detail in all three is untouched.

- **manage_browsers**: names the three ways to drive a session once it exists, in order of preference.
- **execute_playwright_code**: says to use it when a site exposes no WebMCP tools, and to drop to `computer_action` only when the DOM cannot drive the page.
- **computer_action**: says it is the visual last resort and names what to prefer above it.

`webmcp` is unchanged. It was already right.

## What I did not touch

The other 23 tools. Most open with "Manage X" and describe the resource rather than the job, and several already carry a decent "when an agent needs Y" clause. Rewriting all of them is a real piece of work and an unreviewable diff; these four are the ones a model picks between during an actual web task, so they are where the ordering matters.

No test pins this. Asserting on description prose would need either source-text matching or reaching into the SDK's private registry, and both are more brittle than the thing they would protect.

## Testing

`bun test`: 604 pass, 0 fail.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to MCP tool description strings; no runtime, API, or auth behavior changes.
> 
> **Overview**
> Updates the opening sentences of three MCP tool descriptions so models see the same **web execution order** at choice time: **webmcp** → **execute_playwright_code** → **computer_action**, with **manage_browsers** for session lifecycle.
> 
> **`manage_browsers`** now states that sessions underpin every web task and names the three ways to drive a live session, in preference order. **`execute_playwright_code`** is framed as DOM/Playwright (and browser-wide WebMCP helpers) when the site has no own WebMCP tools, with **`computer_action`** only when the DOM cannot drive the page. **`computer_action`** is framed as coordinate/visual control as a last resort, explicitly below site WebMCP and Playwright.
> 
> Operational detail in each description is unchanged; **`webmcp`** and the other tools are not modified in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de860ca094728cc316b62c59940c365e77a05ad9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/mcp/tools/playwright.ts
```diff
@@ -25,7 +25,7 @@ export function registerPlaywrightTool(
   // execute_playwright_code -- Run Playwright/TypeScript code against a browser
   server.tool(
     "execute_playwright_code",
-    "Execute Playwright/TypeScript automation or browser-wide WebMCP helpers against an existing Kernel browser session. Does not create or delete browsers -- use manage_browsers to manage session lifecycle.",
+    "Drive a page through the DOM, using Playwright/TypeScript automation or browser-wide WebMCP helpers, against an existing Kernel browser session. Use this when a site exposes no WebMCP tools of its own, and fall back to computer_action only when the DOM cannot drive the page. Does not create or delete browsers -- use manage_browsers to manage session lifecycle.",
```
masnwilliams, line 28: non-blocking: “Use this when a site exposes no WebMCP tools” is narrower than the actual per-step rule. A site can expose tools and still need Playwright for navigation, inspection, or submitting a non-autosubmit WebMCP form. Could this say to prefer a relevant site tool *for the current step*, then use Playwright when that tool is unavailable or insufficient?

src/lib/mcp/tools/browsers.ts
```diff
@@ -434,7 +434,7 @@ export function registerBrowserCapabilities(
...
+    'Create and manage the browser sessions that every web task runs in, and read their archived telemetry. Once a session exists, drive it with webmcp, execute_playwright_code, or computer_action, in that order of preference. Use "list" to choose an existing session, "create" before browser control, "update" to change supported session settings, "get" for full details, "get_telemetry" to diagnose active or deleted sessions, and "delete" when finished. Live sessions can be addressed by ID or by the name given at creation or set on update; deleted sessions only by ID. get_telemetry compacts events by default; set compact=false with explicit categories and a limit of at most 5 when raw headers, request data, response bodies, or other omitted fields are needed.',
```
masnwilliams, line 437: non-blocking: the new preference order conflicts with `browserSessionNextActions()`, which still recommends `computer_action` first after create/update. Worth aligning that returned hint with this description so the immediate next action does not steer agents to the last resort.

masnwilliams (approved):
This makes the tool-choice order clearer without adding implementation complexity. No blocking code-quality concerns; 604 tests passed locally. Two non-blocking copy alignments below.

